PR #32, opened 2024-03-29
Add runtime tests

drgora:
This PR adds few runtime tests to check the correct configuration of several pallets (e.g. those involved with staking).
It also extends the proofPath E2E test to verify that the expected timeout is respected when publishing a new attestation after the submission of its first proof.

### DIFF
--- a/runtime/src/tests.rs
+++ b/runtime/src/tests.rs
@@ -15,10 +15,30 @@
 
 use super::*;
 
-use frame_support::traits::{fungible::Inspect, Currency, ExistenceRequirement, WithdrawReasons};
+use codec::Encode;
+use frame_support::{
+    assert_ok,
+    traits::{fungible::Inspect, Currency, ExistenceRequirement, OnInitialize, WithdrawReasons},
+};
+use frame_system::{EventRecord, Phase};
 use pallet_settlement_fflonk::{Proof, FULL_PROOF_SIZE};
+use sp_consensus_aura::{Slot, AURA_ENGINE_ID};
+use sp_core::{Pair, Public};
+use sp_runtime::{AccountId32, Digest, DigestItem};
+use sp_staking::{offence, offence::ReportOffence, Exposure, SessionIndex};
 
 mod testsfixtures;
+
+/// Generate a crypto pair from seed.
+pub fn get_from_seed<TPublic: Public>(seed: u8) -> <TPublic::Pair as Pair>::Public {
+    TPublic::Pair::from_string(&format!("//test_seed{}", seed), None)
+        .expect("static values are valid; qed")
+        .public()
+}
+
+pub const SLOT_ID: u64 = 87; // Any random value should do
+pub const NUM_VALIDATORS: u32 = 2;
+pub const AURA_AUTHOR_ID: u32 = (SLOT_ID as u32) % NUM_VALIDATORS;
 
 // Function used for creating the environment for the test.
 // It must return a sp_io::TestExternalities, and the actual test will execute this one before running.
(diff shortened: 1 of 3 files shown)


runtime/src/tests.rs
```diff
@@ -100,3 +168,204 @@ fn pallet_poe_availability() {
...
+    #[test]
+    fn authorship_is_configured_with_aura() {
+        new_test_ext().execute_with(|| {
+            let block_number = 1;
+            let slot = Slot::from(SLOT_ID);
+            let pre_digest = Digest {
+                logs: vec![DigestItem::PreRuntime(AURA_ENGINE_ID, slot.encode())],
+            };
+            System::reset_events();
+            System::initialize(&block_number, &Default::default(), &pre_digest);
```
la10736, line 194: It would be nice put `authorship`'s tests in a `authorship` module and define in this module an `initialize` function that implements this logic and call it for every tests.

runtime/src/tests.rs
```diff
@@ -100,3 +168,204 @@ fn pallet_poe_availability() {
...
+
+            assert!(
```
la10736, line 196: You can use `assert_eq!` here instead

runtime/src/tests.rs
```diff
@@ -100,3 +168,204 @@ fn pallet_poe_availability() {
...
+            )
+            .into_iter()
+            .any(|offender| {
+                pallet_offences::Reports::<Runtime>::get(offender)
+                    .expect("Offence not found")
+                    .offender
+                    .0
+                    == offender_account
+            });
+            assert!(!found);
```
la10736, line 278: I guess here introduce a function like `offenders()` that return an iterators of `AcccountId32` or a more specific `is_offender(AccointId)` function will make this test more readable and simpler to understand.

runtime/src/tests.rs
```diff
@@ -100,3 +168,204 @@ fn pallet_poe_availability() {
...
+                    slash_era: 0,
+                }),
+                topics: vec![],
+            }));
+
+            // Make pallet_offences report an offence
+            let offence = TestOffence {
+                offender_account: offender_account.clone(),
+            };
+            assert!(Offences::report_offence(vec![], offence).is_ok());
```
la10736, line 356: You can use `assert_ok!()` from `frame_support` here.

runtime/src/tests.rs
```diff
@@ -100,3 +168,204 @@ fn pallet_poe_availability() {
...
+            System::reset_events();
+            System::initialize(&block_number, &Default::default(), &pre_digest);
+
+            // Before authoring a block, no points have been given in the active era
+            assert!(
+                Staking::eras_reward_points(Staking::active_era().expect("No active era").index)
+                    .total
+                    == 0
+            );
+
```
la10736, line 242: As general question...  Why do you often check prerequirment contitions in your tests? Almost of the time are trivial true... I'm trying to avoid them and write this checks only when I really need them, but is just mater of tasting if you prefer this form it's fime for me.

la10736 (changes requested):
Great work!!! Just some minor changes.

runtime/src/tests.rs
drgora: Done

runtime/src/tests.rs
```diff
@@ -100,3 +168,204 @@ fn pallet_poe_availability() {
...
+            // Make sure that no slash events for offender_account is published
+            assert!(!System::events().contains(&EventRecord {
+                phase: Phase::Initialization,
+                event: RuntimeEvent::Staking(pallet_staking::Event::SlashReported {
+                    validator: offender_account.clone(),
+                    fraction: Perbill::one(),
+                    slash_era: 0,
+                }),
+                topics: vec![],
+            }));
```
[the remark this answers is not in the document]
drgora: Good point, that was exactly my intention, thanks

runtime/src/tests.rs
drgora: Good idea, done!

runtime/src/tests.rs
```diff
@@ -100,3 +168,204 @@ fn pallet_poe_availability() {
...
+        new_test_ext().execute_with(|| {
+            let session = Session::current_index();
+            let offender_account = AccountId32::new(
+                testsfixtures::SAMPLE_USERS[AURA_AUTHOR_ID as usize]
+                    .raw_account
+                    .into(),
+            );
+
+            // Check that no previous offences were reported
+            let mut found = pallet_offences::ConcurrentReportsIndex::<Runtime>::get(
```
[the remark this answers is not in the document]
drgora: Removed with refactoring

runtime/src/tests.rs
drgora: Done

runtime/src/tests.rs
drgora: Makes sense, done

runtime/src/tests.rs
drgora: I agree it's not a big deal, but I find it helps clarifying the flow of the test and makes the expected precondition stand out.
Also, it should be safer in case of future breaking modifications (even though unlikely)?

a-petrini (approved):
Thanks for this PR! I appreciate the refactoring and improvement of the tests, leaving only a minor non-blocking comment

la10736 (approved):

la10736 (approved):
LGTM

runtime/src/tests.rs
```diff
@@ -15,11 +15,30 @@
...
 mod testsfixtures;
 
+/// Generate a crypto pair from seed.
+pub fn get_from_seed<TPublic: Public>(seed: u8) -> <TPublic::Pair as Pair>::Public {
+    TPublic::Pair::from_string(&format!("//test_seed{}", seed), None)
+        .expect("static values are valid; qed")
+        .public()
+}
+
+pub const SLOT_ID: u64 = 87;
```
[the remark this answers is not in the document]
drgora: Yeah that was just a random value, I added an explicit comment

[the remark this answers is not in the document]
a-petrini: Thanks!

PR #34, opened 2024-04-08
Refactoring of entrypoint.sh

JackPiri:
More flexibility for handling any kind of Substrate params.
The simple rule is based on these examples:

- `NH_CONF_NAME` -> `--name`
- `NH_CONF_BASE_PATH` -> `--base-path`
- `NH_CONF_CHAIN` -> `--chain`

Also removed workaround for handling node-key file.


docker/scripts/entrypoint.sh
```diff
@@ -20,92 +30,51 @@ NH_SECRET_PHRASE_PATH=${NH_SECRET_PHRASE_PATH:-"/data/secret_phrase.dat"}
...
-done
+done < <(env -0)
 
+# Keys handling
 if [ -f "${NH_SECRET_PHRASE_PATH}" ]; then
+  path=""
+  if [ -z "${NH_CONF_BASE_PATH+}" ]; then
+    path=$(get_arg_name_from_env_name NH_CONF_BASE_PATH "${prefix}") $NH_CONF_BASE_PATH
+  fi
+  chain=""
```
a-petrini, line 55: I am not totally familiar of all the use cases this script would be used in, but could there be instances where the lack of explicitly setting `NH_CONF_BASE_PATH` and `NH_CONF_CHAIN` variables would be possible? In this case, and only if this may break things due to unintended behavior, I would add a check that stops the script (or at least warns the user) if either `path` and `chain` local variables result to be empty even after reading them from `env`.

docker/scripts/entrypoint.sh
```diff
@@ -1,6 +1,16 @@
 #!/usr/bin/env bash
 set -eEuo pipefail
 
+get_arg_name_from_env_name() {
```
a-petrini, line 4: I would document a bit this function, or better, I would add a few comments at the beginning of the file explaining which is the syntax of the env variables and how the get converted into `nh-node` parameters, also citing how setting a NH_CONF_XYZ env variable to `"true"` is treated differently (i.e. converted to a argument-less command line option).

a-petrini (changes requested):
Thanks, I left a few comment, and I suggest to improve a bit the documentation of the main script for future reference

docker/scripts/entrypoint.sh
```diff
@@ -20,92 +30,51 @@ NH_SECRET_PHRASE_PATH=${NH_SECRET_PHRASE_PATH:-"/data/secret_phrase.dat"}
...
-  # Check if the variable is defined
-  if [ -n "${var_value}" ]; then
-    echo "${var_name}=${var_value}"
-  else
-    echo "${var_name} is empty"
+prefix="NH_CONF_"
+conf_args=""
+echo "Node configuration:"
+while IFS='=' read -r -d '' var_name var_value; do
+  if [[ "$var_name" == NH_CONF* ]]; then
```
la10736, line 37: Is it possible to use `${prefix}` here?

ysibirski, line 37: I would use `${prefix}` var value here vs `NH_CONF*` explicitly.

la10736 (approved):

docker/scripts/entrypoint.sh
```diff
@@ -20,92 +30,51 @@ NH_SECRET_PHRASE_PATH=${NH_SECRET_PHRASE_PATH:-"/data/secret_phrase.dat"}
...
+    conf_args+="${arg} "
+    echo "  ${var_name}=${var_value} -> $arg"
   fi
-done
+done < <(env -0)
 
+# Keys handling
 if [ -f "${NH_SECRET_PHRASE_PATH}" ]; then
+  path=""
+  if [ -z "${NH_CONF_BASE_PATH+}" ]; then
```
ysibirski, line 52: NO sure about the end goal here, but if NH_CONF_BASE_PATH var is not set next line will produce `unbound variable` error. If it is set to "" empty string `path = --base-path`

Maybe the idea was to ` -n` vs `-z` :thinking:

docker/scripts/entrypoint.sh
ysibirski: Yes, should be allowed.

```
prefix="NH_CONF_"
var_name="NH_CONF_BASE_PATH"

if [[ "${var_name}" == ${prefix}* ]]; then
  echo "FITS"
fi
```

JackPiri: Yes I agree, actually it was that way (`${prefix}*`) but an unintended undo slipped inside :blush:.

JackPiri: Yes I agree, actually it was that way (`${prefix}*`) but an unintended undo slipped inside :blush:.

docker/scripts/entrypoint.sh
JackPiri: The plus sign (`+`) in this parameter expansion does the trick of avoiding unbound variable error, I tested on my machine and it's fine.

docker/scripts/entrypoint.sh
JackPiri: The cases of `NH_CONF_BASE_PATH` and/or `NH_CONF_CHAIN` missing should be fine; btw, in a production environment they would be explicitly defined as environment variables.

docker/scripts/entrypoint.sh
JackPiri: Your suggestion is good, I have added documentation on top of the script.

docker/scripts/entrypoint.sh
a-petrini: Thanks for the clarification!

docker/scripts/entrypoint.sh
JackPiri: @ysibirski I was wrong, what I wrote is not working as expected, please have a look at this fix [commit](https://github.com/HorizenLabs/NH-core/commit/3541fb1376f8a9dc5a811101190c046c4d5406a1) and let me know. Thanks

a-petrini (approved):

docker/scripts/entrypoint.sh
```diff
@@ -12,7 +31,7 @@ get_arg_name_from_env_name() {
 }
 
 # Sanity check
-if [ -z "${BINARY}" ]; then
+if [ -z "${BINARY+}" ]; then
```
ysibirski: This should work but the common used practice to check the var and avoid unbound var error message is
for example 
```
if [ -z "${BINARY:-}" ]; then
```
That means if BINARY is not set then set it to default empty string therefore -z option will produce `true`

However, it is up to you which way to go with. Just a comment :)

docker/scripts/entrypoint.sh
```diff
@@ -16,96 +54,57 @@ IFS=',' read -r -a BINARIES <<< "$BINARY"
...
+    echo "  ${var_name}=${var_value} -> ${arg_name} ${arg_value}"
   fi
-done
+done < <(env -0)
 
+# Keys handling
 if [ -f "${NH_SECRET_PHRASE_PATH}" ]; then
+  injection_args=()
+  path=${NH_CONF_BASE_PATH:-""}
+  if [ -n "${path}" ]; then
```
ysibirski, line 82: You can do this in one line like this
```
if [ -n "${NH_CONF_BASE_PATH:-}" ]; then
```

ysibirski (commented):
See couple of small non blocking comments

docker/scripts/entrypoint.sh
JackPiri: Ok modified accordingly.

docker/scripts/entrypoint.sh
JackPiri: Ok modified accordingly.

docker/scripts/entrypoint.sh
```diff
@@ -16,96 +54,57 @@ IFS=',' read -r -a BINARIES <<< "$BINARY"
...
+    injection_args+=("$(get_arg_name_from_env_name NH_CONF_BASE_PATH ${prefix})")
+    injection_args+=("$(get_arg_value_from_env_value ${NH_CONF_BASE_PATH})")
+  fi
+  chain=${NH_CONF_CHAIN:-""}
+  if [ -n "${chain}" ]; then
```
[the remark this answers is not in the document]
JackPiri: Ok modified accordingly.

PR #35, opened 2024-04-10
Set validators reward to 6.25 NZEN every 2.5 minutes

ptagl:
The current reward mechanism is configured to mimic the coinbase emission on the PoW Horizen blockchain. The only difference is that here the reward is totally assigned to validators, while on Zend a part of the rewards goes to the miners and a small part to the treasury.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -327,14 +327,22 @@ impl pallet_session::Config for Runtime {
 
 pub struct ZendPayout;
 impl pallet_staking::EraPayout<Balance> for ZendPayout {
+    /// Calculates the validators reward based on the duration of the era.
+    /// The reward mimics the current PoW coinbase (6.25 $ZEN per block).
+    /// Total stake and issuance are currently ignored.
     fn era_payout(
         _total_staked: Balance,
         _total_issuance: Balance,
         era_duration_millis: u64,
     ) -> (Balance, Balance) {
-        let era_reward: u128 = 625 * CENTS * u128::from(era_duration_millis / MILLISECS_PER_BLOCK); // 6.25nZEN per block
-        pub const TENTH_TO_MINERS: u128 = 10;
-        pub const TENTH_TO_REST: u128 = 0;
+        const HORIZEN_POW_MILLISECS_PER_BLOCK: u64 = 150 * 1000; // 2.5 minutes
+        let era_reward: u128 =
+            625 * CENTS * u128::from(era_duration_millis / HORIZEN_POW_MILLISECS_PER_BLOCK); // 6.25nZEN per block
+
+        // Assign 100% of the reward to validators.
+        // In the future we may want to split the reward between validators and the treasury.
+        const TENTH_TO_MINERS: u128 = 10;
+        const TENTH_TO_REST: u128 = 0;
 
         (
             era_reward * TENTH_TO_MINERS / 10,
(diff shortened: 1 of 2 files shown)


runtime/src/tests.rs
```diff
@@ -352,3 +352,31 @@ mod pallets_interact {
...
+
+    use crate::{Balance, Runtime, CENTS};
+
+    use super::new_test_ext;
+
+    /// Test that validators receive a cumulative reward that mimics the current emission of
+    /// $ZEN in the PoW Horizen blockchain for miners, which is a coinbase of 6.25 Zen for
+    /// each block every 2.5 minutes.
+    #[test]
+    fn payout_is_same_as_pow_coinbase() {
```
la10736, line 368: Just a mater of taste but I prefer `is_same_as_pow_coinbase`

runtime/src/tests.rs
```diff
@@ -352,3 +352,31 @@ mod pallets_interact {
...
+        new_test_ext().execute_with(|| {
+            const POW_BLOCK_TIME_MILLIS: u64 = 150 * 1000;
+            const POW_BLOCK_COINBASE: Balance = 625 * CENTS;
+            assert_eq!(
+                <Runtime as pallet_staking::Config>::EraPayout::era_payout(
+                    0,
+                    0,
+                    POW_BLOCK_TIME_MILLIS
+                ),
```
la10736, line 377: With just this test you don't have any warrant about your implementation is proportional to the era length. What about add also:

```rust
assert_eq!(<Runtime as pallet_staking::Config>::EraPayout::era_payout(
                    0,
                    0,
                    POW_BLOCK_TIME_MILLIS/10
                ), (POW_BLOCK_COINBASE/10, 0)
);
```

la10736 (approved):
Just a little note. You can decide to ignore them.

runtime/src/lib.rs
```diff
@@ -327,14 +327,22 @@ impl pallet_session::Config for Runtime {
...
-        pub const TENTH_TO_MINERS: u128 = 10;
-        pub const TENTH_TO_REST: u128 = 0;
+        const HORIZEN_POW_MILLISECS_PER_BLOCK: u64 = 150 * 1000; // 2.5 minutes
+        let era_reward: u128 =
+            625 * CENTS * u128::from(era_duration_millis / HORIZEN_POW_MILLISECS_PER_BLOCK); // 6.25nZEN per block
+
+        // Assign 100% of the reward to validators.
+        // In the future we may want to split the reward between validators and the treasury.
+        const TENTH_TO_MINERS: u128 = 10;
+        const TENTH_TO_REST: u128 = 0;
```
JackPiri, line 345: What about adding a static assertion here?
```
// Partition check
static_assertions::const_assert!(TENTH_TO_MINERS + TENTH_TO_REST == 10);
```
Not so useful as of today, but it can be kind of memo for when we extend the number of buckets.

JackPiri (approved):

runtime/src/tests.rs
ptagl: It makes sense, thanks!

runtime/src/lib.rs
ptagl: Good point, I didn't include this check for a matter of time, but it's definitely something we can add in another PR (feel free to open it yourself if you want).

PR #36, opened 2024-04-10
Web3-588 workflow orchestrator and slack notifications

rushby:
This PR introduces a new GitHub Actions orchestrator that runs all the existing jobs, collects artifacts and publishes results to slack via a webhook.

### DIFF
--- a/.github/scripts/summarize_tests.sh
+++ b/.github/scripts/summarize_tests.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+# Set output file locations
+unit_test_file="./unit-test-output/unit_tests_output.txt"
+integration_test_file="./integration-test-output/integration_tests_output.txt"
+coverage_report_file="./coverage-output/coverage_report.json"
+
+# Initialize counters for unit tests
+unit_total_passed=0
+unit_total_failed=0
+unit_total_ignored=0
+unit_total_runtime=0.0
+
+# Initialize counters for integration tests
+integration_total_passed=0
+integration_total_failed=0
+integration_total_ignored=0
+integration_total_runtime=0.0
+
+# Flag for overall status
+overall_status="success"
+
+# Function to process test files (unit or integration)
+process_test_file() {
+  local file_type="$1" # "unit" or "integration"
+  local file_path="$2"
+  
+  if [ -f "$file_path" ]; then
+    while IFS= read -r line; do
+      if [[ "$line" == *"test result:"* ]]; then
+        if [[ "$line" == *"failed; "* && "$overall_status" != "failed" ]]; then
+          overall_status="failed"
+        fi
+
+        passed=$(echo "$line" | grep -oP '(?<=passed; )\d+')
+        failed=$(echo "$line" | grep -oP '(?<=failed; )\d+')
+        ignored=$(echo "$line" | grep -oP '(?<=ignored; )\d+')
+        runtime=$(echo "$line" | grep -oP '(?<=finished in )\d+\.\d+')
+
+        # Update counters based on file type
+        if [[ "$file_type" == "unit" ]]; then
+          unit_total_passed=$((unit_total_passed + passed))
+          unit_total_failed=$((unit_total_failed + failed))
+          unit_total_ignored=$((unit_total_ignored + ignored))
+          unit_total_runtime=$(echo "$unit_total_runtime + $runtime" | bc)
+        elif [[ "$file_type" == "integration" ]]; then
+          integration_total_passed=$((integration_total_passed + passed))
+          integration_total_failed=$((integration_total_failed + failed))
+          integration_total_ignored=$((integration_total_ignored + ignored))
+          integration_total_runtime=$(echo "$integration_total_runtime + $runtime" | bc)
+        fi
+      fi
+    done < "$file_path"
+  else
+    echo "$file_type test output not found."
+  fi
+}
+
+# Process unit and integration test files
+process_test_file "unit" "$unit_test_file"
+process_test_file "integration" "$integration_test_file"
+
+# Extract and summarize overall test coverage data
+if [ -f "$coverage_report_file" ]; then
+    coverage_totals=$(cat "$coverage_report_file" | jq '.data[0].totals')
+    
+    functions_count=$(echo "$coverage_totals" | jq '.functions.count')
+    functions_percent=$(echo "$coverage_totals" | jq '.functions.percent')
+    lines_count=$(echo "$coverage_totals" | jq '.lines.count')
+    lines_percent=$(echo "$coverage_totals" | jq '.lines.percent')
+    regions_count=$(echo "$coverage_totals" | jq '.regions.count')
+    regions_percent=$(echo "$coverage_totals" | jq '.regions.percent')
+    instantiations_count=$(echo "$coverage_totals" | jq '.instantiations.count')
+    instantiations_percent=$(echo "$coverage_totals" | jq '.instantiations.percent')
+
+    coverage_summary="\n*Test Coverage Summary*\n"
+    coverage_summary+="Functions: $functions_count covered, $functions_percent% coverage\n"
+    coverage_summary+="Lines: $lines_count covered, $lines_percent% coverage\n"
+    coverage_summary+="Regions: $regions_count covered, $regions_percent% coverage\n"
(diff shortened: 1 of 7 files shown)


.github/scripts/summarize_tests.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+
```
ysibirski, line 2: Typically, you should be using general set of options for bash scrip like 

```
set -eEuo pipefail
```
It will protect script execution and exit if there is an issue rather then continue bringing misleading results

.github/scripts/summarize_tests.sh
```diff
@@ -0,0 +1,98 @@
...
+    while IFS= read -r line; do
+      if [[ "$line" == *"test result:"* ]]; then
+        if [[ "$line" == *"failed; "* && "$overall_status" != "failed" ]]; then
+          overall_status="failed"
+        fi
+
+        passed=$(echo "$line" | awk '{for (i=1; i<=NF; i++) if ($i=="passed;") print $(i-1)}')
+        failed=$(echo "$line" | awk '{for (i=1; i<=NF; i++) if ($i=="failed;") print $(i-1)}')
+        ignored=$(echo "$line" | awk '{for (i=1; i<=NF; i++) if ($i=="ignored;") print $(i-1)}')
+        runtime=$(echo "$line" | awk -F'finished in ' '{n=split($2,a," "); print (n>0)?a[1]:0}')
```
ysibirski, line 37: Use "" around $() to be safe in case returned value has special chars and/or whitespace(s)

.github/scripts/summarize_tests.sh
```diff
@@ -0,0 +1,98 @@
...
+# Format the run times to 2 decimal places
+unit_total_runtime=$(printf "%.2f" $unit_total_runtime)
+integration_total_runtime=$(printf "%.2f" $integration_total_runtime)
+
+# Format and add summaries to $GITHUB_ENV
+passed_formatted="*Passed*: $unit_total_passed"
+failed_formatted="*Failed*: $unit_total_failed"
+ignored_formatted="*Ignored*: $unit_total_ignored"
+runtime_formatted="*Runtime*: ${unit_total_runtime}s"
+echo "UNIT_TEST_SUMMARY=*Unit Tests*\n$passed_formatted, $failed_formatted, $ignored_formatted, $runtime_formatted" >> $GITHUB_ENV
```
ysibirski, line 71: You need to use `echo -e` in order `\n` to start a new line if that is the goal

.github/scripts/summarize_tests.sh
```diff
@@ -0,0 +1,98 @@
...
+
+passed_formatted="*Passed*: $integration_total_passed"
+failed_formatted="*Failed*: $integration_total_failed"
+ignored_formatted="*Ignored*: $integration_total_ignored"
+runtime_formatted="*Runtime*: ${integration_total_runtime}s"
+echo "INTEGRATION_TEST_SUMMARY=*Integration Tests*\n$passed_formatted, $failed_formatted, $ignored_formatted, $runtime_formatted" >> $GITHUB_ENV
```
ysibirski, line 77: Same here

.github/scripts/summarize_tests.sh
```diff
@@ -0,0 +1,98 @@
...
+    functions_percent=$(echo "$coverage_totals" | jq -r '.functions.percent' | awk '{printf "%.2f", $0}')
+    lines_count=$(echo "$coverage_totals" | jq '.lines.count')
+    lines_percent=$(echo "$coverage_totals" | jq -r '.lines.percent' | awk '{printf "%.2f", $0}')
+    regions_count=$(echo "$coverage_totals" | jq '.regions.count')
+    regions_percent=$(echo "$coverage_totals" | jq -r '.regions.percent' | awk '{printf "%.2f", $0}')
+    instantiations_count=$(echo "$coverage_totals" | jq '.instantiations.count')
+    instantiations_percent=$(echo "$coverage_totals" | jq -r '.instantiations.percent' | awk '{printf "%.2f", $0}')
+
+    coverage_summary="*Test Coverage Summary ($lines_percent%)*\n*Functions:* $functions_count ($functions_percent%), *Lines:* $lines_count ($lines_percent%), *Regions:* $regions_count ($regions_percent%), *Instantiations:* $instantiations_count ($instantiations_percent%)"
+    echo "COVERAGE_TEST_SUMMARY=$coverage_summary" >> $GITHUB_ENV
```
ysibirski, line 93: Same here

ysibirski (commented):
I am not an expert of GH Actions, but see some comments for bash script.

One more comment is just visually orchestrator.yml file is very large. I wonder how hard it will be moving the long if else logic into a script file rather than using inline :shrug:

.github/workflows/CI-orchestrator.yml
```diff
@@ -0,0 +1,170 @@
...
+        run: |
+          if [ -f "./integration-test-output/integration_tests_output.txt" ]; then
+            cat "./integration-test-output/integration_tests_output.txt"
+          else
+            echo "Integration test output not found."
+          fi
+      - name: Display test coverage output
+        run: |
+          if [ -f "./coverage-output/coverage_report.json" ]; then
+            cat "./coverage-output/coverage_report.json"
```
a-petrini, line 57: I don't know if this is possible, but could piping this `cat` with `jq` eventually improve readability of the report?

a-petrini (approved):
Thanks for this PR! It looks very good and I just left a minor non-blocking comment.

.github/workflows/CI-orchestrator.yml
```diff
@@ -0,0 +1,170 @@
+name: Orchestrator
+
+on:
+  pull_request:
```
JackPiri, line 4: Probably we want to keep previous trigger:
```
on:
  pull_request:
    types: [opened, synchronize]
```
in order to avoid for example the orchestrator to run on `pull_request.updated` event (description changed, title changed or other non-code updates).

.github/workflows/CI-build-test.yml
```diff
@@ -47,6 +43,27 @@ jobs:
...
-        run: cargo test --test '*' --all-features --no-fail-fast --release
+        run: |
+          cargo test --test '*' --all-features --no-fail-fast --release 2>&1 | tee integration_tests_output.txt
+      - name: Upload Unit Test Output
+        uses: actions/upload-artifact@v4
+        with:
+          name: unit-test-output
+          path: unit_tests_output.txt
+          if-no-files-found: warn
+          retention-days: 1
```
JackPiri, line 57: Is `retention-days: 1` actually working as expected? I was able to download the file today, much more than 1 day has passed since the target run.

.github/workflows/CI-lint-format.yml
```diff
@@ -43,8 +39,24 @@ jobs:
...
+          path: linting_output.txt
+          if-no-files-found: warn
+          retention-days: 1
+          compression-level: 0
+          overwrite: true
       - name: Formatting
-        if: matrix.os == 'ubuntu-latest'
-        run: cargo fmt --check
\ No newline at end of file
+        run: cargo fmt -- --check 2>&1 | tee formatting_output.txt
```
JackPiri, line 53: Is that additional `--` introduced on purpose or is it a leftover? It seems to me no big difference btw.

JackPiri (approved):
The PR is fine to me. I've left some minor comments.
Thanks

.github/workflows/CI-build-test.yml
rushby: Interesting! I believe it's implemented as per the docs, but it sounds like something isn't quite right.  I'll keep an eye on it!

.github/scripts/summarize_tests.sh
rushby: Thanks - amended!

.github/workflows/CI-orchestrator.yml
rushby: Thanks Alessandro! It's a good question, maybe it is an improvement for the future if required

.github/workflows/CI-orchestrator.yml
rushby: Thanks for this, good spot!

.github/workflows/CI-lint-format.yml
rushby: Good spot i've removed that additional --

PR #37, opened 2024-04-18
Reviewed docker file and gh actions ci

ysibirski:
@lander86 I have checked all the dockerfiles and applied small modifications simplifications as well as push to docker hub versioned tag.

However, we should not be making this repo public until we prevent people to be able to trigger GH Actions run and image build. We may want to introduce **MAINTAINER_KEYs**  variable and **setup_env.sh** approach as we do it in other places to be safe, otherwise if someone will be able to build and publish new image we can get in trouble.

Also with the current logic, tests are running `cargo build --release` as well as docker image build pipeline. It is time consuming process and I think there will be a good benefit to add cargo cache, therefore you need to run it only once and share artifacts between jobs


rushby:
Agreed about caching, just had a similar discussion.

.github/workflows/CI-docker.yml
```diff
@@ -4,42 +4,40 @@ run-name: "Workflow performing CI step: push docker image to DockerHub"
...
 
 env:
   RUST_BACKTRACE: 1
+  DOCKER_HUB_ORG: horizenlabs
+  DOCKER_IMAGE_NAME: nh-node
+  PROD_VERSION_REGEX: '^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9]+)?(-rc[0-9]+)?$'
 
 jobs:
   docker:
     strategy:
```
JackPiri, line 18: I think we can take the chance to simplify this from:
```
    strategy:
      matrix:
        os: [ubuntu-latest]
    runs-on: ${{ matrix.os }}
```
to
```
    runs-on: ubuntu-latest
```

.github/workflows/CI-docker.yml
```diff
@@ -4,42 +4,40 @@ run-name: "Workflow performing CI step: push docker image to DockerHub"
...
     tags:
-      - v[0-9].[0-9].[0-9]*
+      - 'v[0-9].[0-9].[0-9]*'
   workflow_dispatch:
 
 env:
   RUST_BACKTRACE: 1
+  DOCKER_HUB_ORG: horizenlabs
+  DOCKER_IMAGE_NAME: nh-node
+  PROD_VERSION_REGEX: '^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9]+)?(-rc[0-9]+)?$'
```
JackPiri, line 14: I had a brief discussion with Luigi on this, I am not totally fine with this `PROD_VERSION_REGEX` logic.
Just to make sure we are aligned on what was our previous approach:

- `v1.0.0`: fine, tag `1.0.0` published, tag `latest` published
-` v1.0.0-rc.1`: fine, `1.0.0-rc.1` published, tag `latest` NOT published (because not an official/final version)
-` v1.0.0-beta.1`: fine, `1.0.0-beta.1` published, tag `latest` NOT published (as above)
- `v1.0.0-blablabla`: fine, `1.0.0-blablabla` published, tag `latest` NOT published (as above); maybe not super-good but acceptable
What I don't like very much is that with your new approach `v1.0.0-rc.1` would be published as `latest` (can be a problem for someone joining the network after the publishing of `v1.0.0-rc.1` but before the publishing of `v1.0.0`) and `v1.0.0-beta.1` would be rejected at all (but maybe we want to test it on some dedicated machines).

.github/workflows/CI-docker.yml
```diff
@@ -4,42 +4,40 @@ run-name: "Workflow performing CI step: push docker image to DockerHub"
...
 
 jobs:
   docker:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        # os: [ubuntu-latest, windows-latest, macos-latest]
```
JackPiri, line 18: I think we can take the chance to simplify this from:
```
    strategy:
      matrix:
        os: [ubuntu-latest]
    runs-on: ${{ matrix.os }}
```
to
```
    runs-on: ubuntu-latest
```

docker/dockerfiles/hl-node.Dockerfile
```diff
@@ -19,53 +19,40 @@ FROM ubuntu:22.04 as node
...
-
-ARG BIN_FOLDER=.
-ARG DOC_URL=https://github.com/HorizenLabs/NH-core
-ARG DESCRIPTION="NH-core"
-ARG AUTHORS="devops@horizenlabs.io"
+ARG DESCRIPTION="New Horizen Core"
+ARG AUTHORS="mainchain-team@horizenlabs.io"
 ARG VENDOR="Horizen Labs"
 
-LABEL io.hl.image.authors=${AUTHORS} \
```
JackPiri, line 36: I agree on removing label `revision`, not sure on removing `created` and `documentation`.

JackPiri (changes requested):
As of now external contributor are not allowed to push a tag on remote (tried on a VM without stored credentials and got request for usr+pwd prompted); but please double check for confirmation.
Regarding the PR, please take a look at the comment related to `PROD_VERSION_REGEX` and also address the missing removal of letter `v` from GitHub-tag to Docker-tag.

.github/workflows/CI-docker.yml
ysibirski: If we are planning to drop `v` in the tag name, why not to drop it as well for github tag name as well? use 1.1.1 for github tag vs v1.1.1?

How about we define a fixed pattern for allowed tags to be published like ONLY
1) If 1.1.1 = PRODUCTION release tags are 1.1.1 and latest are published
2) If 1.1.1-rc[1-9] = DEV release and ONLY 1.1.1-rc[1-9] is published

Why do we need beta and smth else?
We can always add it if needed of course

.github/workflows/CI-docker.yml
```diff
@@ -4,42 +4,40 @@ run-name: "Workflow performing CI step: push docker image to DockerHub"
...
-          if [[ "${TAG_NAME}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-            docker image tag nh-node:build horizenlabs/nh-node:latest
-            docker push horizenlabs/nh-node:latest
-          else
-            echo Not publishing latest because not official version
-          fi
-      - name: Logout from DockerHub
-        run: docker logout
+          docker login --username ${{ secrets.DOCKER_HUB_USERNAME }} --password ${{ secrets.DOCKER_HUB_TOKEN }}
+          DOCKER_IMAGE_VERSION_TAG="${{ github.ref_name }}"
```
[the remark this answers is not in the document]
ysibirski: The questoin here if we want then to use `v` in github tag name?

docker/dockerfiles/hl-node.Dockerfile
ysibirski: My approach was to reduce as many moving parts as possible, so we are not going to be dependent on git repo name and etc in case it changes one day. AFAIK labels are not necessary, but I am ok to keep whatever you guys think MUST be kept. Let me know what do you guys think. FYI: we are not creating any labels like this for either zend or eon IIRC

cc: @lander86

ysibirski:
@lander86 and @JackPiri I have pushed new version of the pipeline, but some questions still need to be answered like for example github tag name vs docker tag version and labels

.github/workflows/CI-docker.yml
lander86: Let's drop the `v` in GH.
About the tagging patterns I like your proposal @ysibirski but I would also allow "free tags" like : 

If 1.1.1 = PRODUCTION release tags are 1.1.1 and latest are published can ONLY be done if tag is from MAIN branch
If 1.1.1-rc[1-9] = DEV release and ONLY 1.1.1-rc[1-9] is published can be done from ANY branch
If 1.2.0-anyword = only 1.2.0-anyword tag get published on DH. 

It can be useful for MC team to be able to push also specific experimental tags. 
We will anyway have the check on the maintainers keys.

JackPiri: Ok we are perfectly aligned, thanks.

lander86:
We still miss the CODEOWNERS file, the new strategy to run the CI and to align the branch with main (we're missing some changes recently merged and needed to push some updates)

lander86 (approved):
LGTM

.github/workflows/CI-docker.yml
```diff
@@ -1,45 +1,30 @@
...
-        run: |
-          docker build \
-          --build-arg VCS_REF="${VCS_REF}" \
-          --build-arg BUILD_DATE="$(date -u '+%Y-%m-%dT%H:%M:%SZ')" \
-          --build-arg IMAGE_NAME="hl/nh-node" \
-          -f docker/dockerfiles/hl-node.Dockerfile -t nh-node:build .
-      - name: Login into DockerHub
-        run: docker login --username cihorizen --password ${{ secrets.DOCKER_HUB_TOKEN }}
-      - name: Publish Docker image(s) to DockerHub
+        with:
```
JackPiri, line 20: I see no need for this:
```
        with:
          ref: ${{ github.ref }}
```

ci/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+export PROD_RELEASE="false"
+export DEV_RELEASE="false"
+export TEST_RELEASE="false"
+export COMMON_FILE_LOCATION='ci/common.sh'
+
+workdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
+github_tag="${GITHUB_REF_NAME:-}"
+prod_release_branch="${PROD_RELEASE_BRANCH:-main}"
+prod_release_regex='^[0-9]+\.[0-9]+\.[0-9]+$'
+dev_release_regex='^[0-9]+\.[0-9]+\.[0-9]+(-[0-9]+)?(-rc[0-9]+){1}$'
```
JackPiri, line 14: Please modify this as:
dev_release_regex='^[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+){1}$'
We want `1.2.3-rc.1` to match, we don't want `1.2.3-rc1` nor `1.2.3-1234-rc.1` to match.

ci/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+test_release_regex='^[0-9]+\.[0-9]+\.[0-9]+(-[0-9]+)?-.*$'
```
JackPiri, line 15: Please modify this as:
`test_release_regex='^[0-9]+\.[0-9]+\.[0-9]+.+$'`.
Would make things easier and distinguish between `prod_release_regex` case.

ci/setup_env.sh
```diff
@@ -0,0 +1,114 @@
...
+      if [[ "${github_tag}" =~ ${dev_release_regex} ]]; then
+        export DEV_RELEASE="true"
+      elif [[ "${github_tag}" =~ ${test_release_regex} ]]; then
+        export TEST_RELEASE="true"
+      else
+        log bold yellow "WARNING: GitHub tag: ${github_tag} is in the wrong format for the DEVELOPMENT or TEST release. Expecting the following format of the release: DEVELOPMENT = 'd.d.d-rc[0-9]' and TEST = 'd.d.d-.*'. The build is not going to be released ..."
+        export IS_A_RELEASE="false"
+      fi
+    fi
+  fi
```
JackPiri, line 100: Please rewrite this as below:
```
  # Release test
  if [ "${IS_A_RELEASE}" = "true" ]; then
    if (git branch -r --contains "${github_tag}" | grep -xqE ". origin\/${prod_release_branch}$"); then
      # Checking format of production vs development release version
      if [[ "${github_tag}" =~ ${prod_release_regex} ]]; then
        export PROD_RELEASE="true"
      elif [[ "${github_tag}" =~ ${dev_release_regex} ]]; then
        export DEV_RELEASE="true"
      elif [[ "${github_tag}" =~ ${test_release_regex} ]]; then
        export TEST_RELEASE="true"
      else
        log bold yellow "WARNING: GitHub tag: ${github_tag} is in the wrong format for any type of release (PRODUCTION, DEVELOPMENT, TEST). The build is not going to be released ..."
        export IS_A_RELEASE="false"
      fi
    else
      if [[ "${github_tag}" =~ ${prod_release_regex} ]]; then
        log bold yellow "WARNING: GitHub tag = ${github_tag} is in the PRODUCTION format but was not created from PRODUCTION branch = ${prod_release_branch}."
        export IS_A_RELEASE="false"
      elif [[ "${github_tag}" =~ ${dev_release_regex} ]]; then
        export DEV_RELEASE="true"
      elif [[ "${github_tag}" =~ ${test_release_regex} ]]; then
        export TEST_RELEASE="true"
      else
        log bold yellow "WARNING: GitHub tag: ${github_tag} is in the wrong format for the DEVELOPMENT or TEST release. The build is not going to be released ..."
        export IS_A_RELEASE="false"
      fi
    fi
  fi
```
Apart from logs adaptation, the main point here is that we could want (and actually almost always do this way) to create tags on `main`, also for dev/test versions.

ci/docker.sh
```diff
@@ -0,0 +1,70 @@
...
+
+docker_tag=""
+if [ "${is_a_release}" = "true" ]; then
+  docker_tag="${github_ref_name}"
+fi
+
+# Building and publishing docker image
+if [ -n "${docker_tag:-}" ]; then
+  log italic green "=== Building Docker image: ${docker_hub_org}/${docker_image_build_name}:${docker_tag} ==="
+  #docker build -f "${docker_file_path}" -t "${docker_hub_org}/${docker_image_build_name}:${docker_tag}" .
```
JackPiri, line 46: Please uncomment `docker build`, `docker tag` and `docker push` commands.

JackPiri (changes requested):
Please address comments for prod/dev/test labelling and flows, also uncomment `docker` commands.

.github/workflows/CI-docker.yml
ysibirski: This is actually a very important one to have. The pipeline will not work correctly without it. Took me couple hours to figure this out :face_exhaling: . Long story shout without this `ref` Github Tag will be checked out by its commit hash vs tag therefore `git verify-tag` command that checks if git tag is signed by the maintainers_key will fail since it ONLY works on the object = tag and not object = commit.

We should keep this in mind for the future cases like this.

ci/setup_env.sh
ysibirski: From our conversation with @lander86 we agreed on implementing the common nomenclature for releases which is `d.d.d-rc[0-9]*` 

I think we need to meet all together and come up with the solution that everyone agrees on.

ci/setup_env.sh
ysibirski: Don't we need `-` as well? like `d.d.d-ANYTHING` ?

ci/docker.sh
ysibirski: That was the plan for the last commit once everything else is reviewed and approved. The plan is not to build docker image everytime during the implementation

ci/setup_env.sh
lander86: @JackPiri I would keep the format like N.N.N-Anything

JackPiri: Ok it's fine to me.

ci/setup_env.sh
ysibirski: Spoke with @lander86 about it and seems like we should not be releasing anything from main rather than PROD and should use development branch or smth like this for this kind of scenario

JackPiri (approved):
The PR is fine to me.
Thanks

PR #38, opened 2024-04-19
Babe integration

drgora:
This PR updates the consensus to drop Aura block authoring in favor of Babe.

Runtime tests are updated and integrated to account for the new pallet, and its interactions.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -56,11 +56,7 @@ fn from_ss58check<T: sp_core::crypto::Ss58Codec>(
     <T as sp_core::crypto::Ss58Codec>::from_ss58check(key)
 }
 
-fn session_keys(
-    babe: BabeId,
-    grandpa: GrandpaId,
-    im_online: ImOnlineId,
-) -> SessionKeys {
+fn session_keys(babe: BabeId, grandpa: GrandpaId, im_online: ImOnlineId) -> SessionKeys {
     SessionKeys {
         babe,
         grandpa,
@@ -150,9 +146,9 @@ pub fn local_config() -> Result<ChainSpec, String> {
         WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?,
         None,
     )
-    .with_name("DR NH Local")
-    .with_id("dr_nh_local")
-    .with_protocol_id("dr_nh_local")
+    .with_name("NH Local")
+    .with_id("nh_local")
+    .with_protocol_id("lzen")
     .with_chain_type(ChainType::Local)
     .with_properties({
         let mut props = Properties::new();
@@ -186,6 +182,7 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
     )
     .with_name("NH Testnet")
     .with_id("nh_testnet")
+    .with_protocol_id("tzen")
     .with_chain_type(ChainType::Live)
     .with_properties({
         let mut props = Properties::new();
@@ -305,7 +302,6 @@ fn genesis(
             "balances": endowed_accounts,
         },
         "babe": {
-            //authorities: vec![],
             "epochConfig": Some(nh_runtime::BABE_GENESIS_EPOCH_CONFIG),
         },
         "session": {
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -18,6 +18,7 @@ use nh_runtime::{currency, AccountId, RuntimeGenesisConfig, SessionKeys, Signatu
 use pallet_im_online::sr25519::AuthorityId as ImOnlineId;
 use sc_service::{ChainType, Properties};
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
+use sp_consensus_babe::AuthorityId as BabeId;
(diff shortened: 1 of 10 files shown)


runtime/src/tests.rs
```diff
@@ -173,37 +189,72 @@ fn pallet_poe_availability() {
...
+    // Any random values for these constants should do
+    const BLOCK_NUMBER: BlockNumber = 1;
+    const SLOT_ID: u64 = 87;
+    const BABE_AUTHOR_ID: u32 = 1;
+
+    // Initialize block #BLOCK_NUMBER, authored at slot SLOT_ID by BABE_AUTHOR_ID using Babe
+    fn initialize() {
+        let slot = Slot::from(SLOT_ID);
+        let authority_index = BABE_AUTHOR_ID;
+        let transcript = sp_consensus_babe::VrfTranscript::new(b"test", &[]);
```
la10736, line 201: You're using this function to init the chain in almost all following tests. Consider to redefine `new_test_ex()` function to include this code too:

```rust
    fn new_test_ext() -> sp_io::TestExternalities {
        let mut ex = super::new_test_ext();
        ex.execute_with(|| initialize());
        ex
    }
```

node/src/chain_spec.rs
```diff
@@ -148,6 +148,7 @@ pub fn local_config() -> Result<ChainSpec, String> {
     )
     .with_name("NH Local")
     .with_id("nh_local")
+    .with_protocol_id("nh_local")
```
la10736, line 151: We can use `zen` or `tzen` or `nzen` .... seams more close to what polkadot do

node/src/chain_spec.rs
```diff
@@ -181,6 +182,7 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
     )
     .with_name("NH Testnet")
     .with_id("nh_testnet")
+    .with_protocol_id("nh_testnet")
```
la10736, line 185: Here we can use `tzen`

la10736 (approved):
LGTM

runtime/src/tests.rs
drgora: Good idea, done!

node/src/chain_spec.rs
drgora: Done

node/src/chain_spec.rs
drgora: Done

JackPiri:
Please modify entrypoint.sh as suggested below:
```
  echo "Injecting key (Babe)"
  ${NH_NODE} key insert "${injection_args[@]}" \
    --scheme Sr25519 \
    --suri "${NH_SECRET_PHRASE_PATH}" \
    --key-type babe
```
(reference: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/core/src/crypto.rs)

runtime/src/tests.rs
```diff
@@ -18,27 +18,35 @@ use super::*;
...
     TPublic::Pair::from_string(&format!("//test_seed{}", seed), None)
         .expect("static values are valid; qed")
-        .public()
 }
 
-pub const SLOT_ID: u64 = 87; // Any random value should do
-pub const NUM_VALIDATORS: u32 = 2;
-pub const AURA_AUTHOR_ID: u32 = (SLOT_ID as u32) % NUM_VALIDATORS;
+const NUM_VALIDATORS: u32 = 2;
+const TEST_PRIMARY_PROBABILITY: (u64, u64) = (1, 4);
```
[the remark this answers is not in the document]
drgora: Yes, makes sense, done

runtime/src/lib.rs
```diff
@@ -352,7 +363,7 @@ impl pallet_staking::EraPayout<Balance> for ZendPayout {
 }
 
 parameter_types! {
-    pub const SessionsPerEra: sp_staking::SessionIndex = 6 * HOURS / Period::get(); // number of sessions in 1 era, 6h
+    pub const SessionsPerEra: sp_staking::SessionIndex = 6 * HOURS / EpochDurationInBlocks::get(); // number of sessions in 1 era, 6h
```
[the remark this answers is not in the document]
drgora: That would be fine and equivalent in terms of era duration, but longer sessions might incur longer response times in session handling (for example if active validators misbehave and a new set should be activated).

drgora:
> Please modify `entrypoint.sh` as suggested below:
> 
> ```
>   echo "Injecting key (Babe)"
>   ${NH_NODE} key insert "${injection_args[@]}" \
>     --scheme Sr25519 \
>     --suri "${NH_SECRET_PHRASE_PATH}" \
>     --key-type babe
> ```
> 
> (reference: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/core/src/crypto.rs)

Done as suggested, thank you.

ptagl (approved):

runtime/src/lib.rs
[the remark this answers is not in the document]
JackPiri: Ok we are fine in increasing the session length from 100 blocks to 600 blocks.

PR #39, opened 2024-04-22
CI rustdoc generation

ptagl:
This PR introduces a new GitHub Actions workflow to automatically generate the Rust documentation through `cargo doc` and then push it to GitHub Pages.

For simplified navigation, the documentation homepage is configured to redirect to the `nh_node` crate homepage.

Additionally, the workflow used for the Docker image build and push is now bundled into a "tag" workflow that includes the documentation as well.

### DIFF
--- a/.github/workflows/CI-rustdoc.yml
+++ b/.github/workflows/CI-rustdoc.yml
@@ -26,8 +26,8 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-lint-format-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: ${{ runner.os }}-lint-format-cargo-
+          key: ${{ runner.os }}-build-test-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-build-test-cargo-
 
       - name: Install deps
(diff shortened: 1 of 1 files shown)


a-petrini (approved):
Looks good, thanks for this PR!

.github/workflows/CI-rustdoc.yml
```diff
@@ -0,0 +1,71 @@
...
+        uses: actions/cache@v4
+        continue-on-error: false
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-lint-format-cargo-${{ hashFiles('**/Cargo.lock') }}
```
JackPiri, line 29: Is `lint-format` cache here used on purpose? My feeling would be to use `build-test` cache, but if your is a motivated choice just ignore this comment.

JackPiri (approved):
The PR is fine to me.
Thanks

.github/workflows/CI-rustdoc.yml
ptagl: It makes sense to me, thanks! I've also updated the `cargo doc` command adding the `--release` flag as the cache is built with the same flag.

PR #44, opened 2024-04-29
Test E2E that's check both valid and fail proofs should pay

la10736:
1. Tests are listed by search in the folder
2. A new E2E that checks both valid and fail proof pay.

### DIFF
--- a/e2e-tests/js_scripts/0006-proof_should_pay.js
+++ b/e2e-tests/js_scripts/0006-proof_should_pay.js
@@ -1,0 +1,138 @@
+// This script is used to test the proof verification pay a fee both if it's ok or fails.
+
+const Keccak256 = require('keccak256')
+
+// Hardcoded proof hashes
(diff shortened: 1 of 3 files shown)


drgora (approved):
I believe we should refactor e2e tests to create a `util` library to avoid code duplication, but we can do this in a future PR.

LGTM, thanks!

ptagl (approved):

a-petrini (commented):
Looks good, thanks for this PR!

e2e-tests/js_scripts/0006-proof_should_pay.js
```diff
@@ -0,0 +1,138 @@
...
+
+const MIN_PRICE = 1000000000;
```
JackPiri, line 9: What's the rationale behind value of `MIN_PRICE`? I would have set this to `1 `(or just left the assertion something like `balanceBefore<balanceAfter`), but maybe there is something I'm missing.

la10736: I just used something that isn't trivial but sensible lesser than the one on the weight. I don't want both to strongly couple it to the used weight and to use a trivial value that is comparable to an nop extrinsic.

JackPiri: Ok thanks

PR #46, opened 2024-05-03
Md/zksync-verifier

la10736:
Introduce zksync verifier pallet. In this case we need to use a native implementation (introduce this pallet is not a forkless upgrade)

We covered both
- pallet unit tests
- runtime inclusion test
- e2e tests via replacing a one of two fflonk proofs with a zksync one.

Also introduce `script/bench.sh` to compute weight benchmark


scripts/bench.sh
```diff
@@ -0,0 +1,65 @@
...
+
+PALLET=$1
+WEIGTH_TEMPLATE=${WEIGTH_TEMPLATE:-"${DEFAULT_WEIGTH_TEMPLATE}"}
+WEIGTH_OUT_PATH=${WEIGTH_OUT_PATH:-""}
+
+function usage {
+    local message=${1:-""};
+
+    echo "$0 <pallet> : get pallet crate name, execute benchamark and save the weigth file.
+    Evirorments:
```
drgora, line 13: Small typo here:
```suggestion
    Environment:
```

scripts/bench.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+DEFAULT_WEIGTH_TEMPLATE="node/frame-weight-template.hbs"
```
drgora, line 3: Small typo here (and below): `WEIGTH` -> `WEIGHT`

pallets/settlement-zksync/src/tests.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        // Dispatch a signed extrinsic.
+        assert_noop!(
+            SettlementZksyncPallet::submit_proof(RuntimeOrigin::signed(1), data.into()),
+            expected_error
+        );
+    })
+}
+
+#[rstest]
```
drgora, line 73: Does this actually need to be an `rstest` and not a `test`? It looks like you are not using any rstest feature here.

drgora (changes requested):
Looks great to me! I believe it only needs the `zksync-era-verifier` dependency to be fixed for being approved.

native/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+description = "Provide a native implementation bridge for wasm executor"
+version = "0.1.0"
+edition.workspace = true
+authors.workspace = true
+repository.workspace = true
+homepage.workspace = true
+
+[dependencies]
+zksync-era-verifier = { git = "https://github.com/HorizenLabs/zksync-era-verifier.git", branch = "first-commit", optional = true }
+zksync-era-verifier-deserialize = { git = "https://github.com/HorizenLabs/zksync-era-verifier.git", branch = "first-commit", optional = true }
```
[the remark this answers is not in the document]
la10736: Right!!! thx

pallets/settlement-zksync/src/tests.rs
la10736: I had changed it in a experiment whild I was using a fixture for tests excutor... later I rollbacked it and forget the `#[rstest]` annotation. Anyway it doesn't introduce any issue but of cource is useless

drgora (approved):

pallets/settlement-zksync/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub use native::ZKSYNC_PROOF_SIZE as PROOF_SIZE;
+pub use native::ZKSYNC_PUBS_SIZE as PUBS_SIZE;
+pub const FULL_PROOF_SIZE: usize = PROOF_SIZE + PUBS_SIZE;
```
JackPiri, line 20: I like this and make me think we can also change fflonk side from:
`pub const FULL_PROOF_SIZE: usize = 25 * 32;`
to
`pub const FULL_PROOF_SIZE: usize = PROOF_SIZE + PUBS_SIZE;`

pallets/settlement-zksync/src/tests.rs
```diff
@@ -0,0 +1,98 @@
...
+        // Dispatch a signed extrinsic.
+        assert_noop!(
+            SettlementZksyncPallet::submit_proof(RuntimeOrigin::signed(1), data.into()),
+            expected_error
+        );
+    })
+}
+
+#[test]
+fn invalid_proof_fails_verification_and_is_not_notified() {
```
JackPiri, line 74: It seems to me test `invalid_proof_fails_verification_and_is_not_notified` is already covered by `malformed_data_fails_verification_and_is_not_notified` with `case::invalid_pubs`; so I propose to remove test `invalid_proof_fails_verification_and_is_not_notified`.

pallets/settlement-zksync/src/tests.rs
```diff
@@ -0,0 +1,98 @@
...
+            .into(),
+        );
+    });
+}
+
+#[rstest]
+#[case::syntactic_valid_proof((730, 0x00), Error::<Test>::VerifyError)]
+#[case::syntactic_invalid_proof((0, 0x00), Error::<Test>::InvalidProofData)]
+#[case::invalid_pubs((1408, 0xFF), Error::<Test>::InvalidInput)]
+fn malformed_data_fails_verification_and_is_not_notified(
```
JackPiri, line 57: Please rename this test from `malformed_data_fails_verification_and_is_not_notified` to `malformed_data_fails_verification`, since we are not checking the notification is missing.

JackPiri:
Just a general consideration: maybe we can think about moving some code from verifier pallets to a primitives module, I am thinking for example to `compute_hash` function (being almost identically implemented in both fflonk-verifier and zksync-verifier), error code (`InvalidInput`, `InvalidProofData`, `VerifyError` which seem a kind of standard pattern for any verifier).
Can be done here or also in a later PR.

pallets/settlement-zksync/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+                .map_err(Into::into)
+        }
+    }
+
+    const PREFIX: &[u8; 7] = b"zksync-";
+    fn compute_hash(pubs: &[u8]) -> H256 {
+        let mut data_to_hash = PREFIX.to_vec();
+        data_to_hash.extend_from_slice(&pubs);
+        H256(keccak_256(data_to_hash.as_slice()))
+    }
```
JackPiri, line 142: Not directly related to your changes, but I see test `fflonk_hash_as_expected` can be removed, being already covered by `valid_proof_passes_verification_and_is_notified` (`pallets/settlement-fflonk/src/tests.rs`).

JackPiri (approved):
The PR is fine to me. I've left some minor/optional comments.
Thanks

la10736:
> Just a general consideration: maybe we can think about moving some code from verifier pallets to a primitives module, I am thinking for example to `compute_hash` function (being almost identically implemented in both fflonk-verifier and zksync-verifier), error code (`InvalidInput`, `InvalidProofData`, `VerifyError` which seem a kind of standard pattern for any verifier). Can be done here or also in a later PR.

Right!
I did the same thought when I wrote it. But my point here is to wait the third duplication before do a refactoring.

pallets/settlement-zksync/src/lib.rs
la10736: Maybe is better to cover it in a new PR

pallets/settlement-zksync/src/tests.rs
la10736: No, it isn't. `malformed_data_fails_verification_and_is_not_notified::invalid_pubs` checks what happens when the public input are invalid (error type `InvalidInput`) and `invalid_proof_fails_verification_and_is_not_notified` is about a use a valid input type but that value isn't the valid public input for this proof (error type `VerifyError`).

Anyway I'll move this test in the previous parametric group.

pallets/settlement-zksync/src/lib.rs
la10736: A later PR is beter

PR #47, opened 2024-05-03
This readme is empty and incoherent: remove it

la10736:


JackPiri (commented):
In a similar way we probably want to remove also `pallets/proof_of_existence/README.md`, correct?

la10736:
> In a similar way we probably want to remove also `pallets/proof_of_existence/README.md`, correct?

Right!

JackPiri (approved):
The PR is fine to me.
Thanks

ptagl (approved):

PR #48, opened 2024-05-08
Gp/bootnodes chain spec

JackPiri:
Modifications for chain spec (testnet only):

- added bootnodes connection strings
- added telemetry endpoint

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -22,6 +22,10 @@ use sp_consensus_grandpa::AuthorityId as GrandpaId;
 use sp_core::{sr25519, Pair, Public};
 use sp_runtime::traits::{IdentifyAccount, Verify};
 
+// The connection strings for bootnodes
+const BOOTNODE_1: &str = "/dns/bootnode-tn-1.zkverify.io/tcp/443/wss/p2p/12D3KooWNhvf6iSowraUY4tZnjpNZXEe85oy9zDWYRKFBnWivukc";
+const BOOTNODE_2: &str = "/dns/bootnode-tn-2.zkverify.io/tcp/443/wss/p2p/12D3KooWEjVadU1YWyfDGvyRXPbCq2rXhzJtXaG4RxJZBkGE9Aug";
+
 // The URL for the telemetry server.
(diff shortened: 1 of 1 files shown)


node/src/chain_spec.rs
```diff
@@ -184,6 +185,18 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
...
+    .with_boot_nodes(vec![
+        "/dns/nh-bootnode-1.horizenlabs.io/tcp/443/wss/p2p/12D3KooWNhvf6iSowraUY4tZnjpNZXEe85oy9zDWYRKFBnWivukc"
+        .parse()
+        .expect("MultiaddrWithPeerId"),
+        "/dns/nh-bootnode-2.horizenlabs.io/tcp/443/wss/p2p/12D3KooWEjVadU1YWyfDGvyRXPbCq2rXhzJtXaG4RxJZBkGE9Aug"
+        .parse()
+        .expect("MultiaddrWithPeerId"),
+    ])
+    .with_telemetry_endpoints(
+		TelemetryEndpoints::new(vec![("wss://nh-telemetry.horizenlabs.io/submit".to_string(), 1)])
```
drgora, line 197: I'd suggest replacing the verbosity level value with its named constant.
```suggestion
		TelemetryEndpoints::new(vec![("wss://nh-telemetry.horizenlabs.io/submit".to_string(), sc_telemetry::CONSENSUS_INFO)])
```

drgora (commented):
Another note: I believe we should either remove the `STAGING_TELEMETRY_URL` (currently commented out), or actually use it with our endpoint in place of the raw value at line 197.

la10736 (approved):
LGTM

node/src/chain_spec.rs
JackPiri: Ok I agree on both notes.
I have replaced `1` with `sc_telemetry::CONSENSUS_INFO` and now `STAGING_TELEMETRY_URL` is set and used (done similarly also for bootnodes connection strings).

drgora (approved):
Looks good to me

lander86 (approved):
LGTM

PR #49, opened 2024-05-09
Changed ticker name. Updated balances and keys. Introduced inflation curve for era payouts.

DanieleDiBenedetto:
As per title.
Inflation curve parameters have been taken from Polkadot, we'd need to customize them according to our needs in future releases.
All the code and tests related to the simulation of ZEND payouts have been removed.
CDK aggregator address has been removed as well from the chainspec.
Ticker name has been changed to ZETA.
Balances have been updated to sum to around 1B

### DIFF
--- a/runtime/src/tests.rs
+++ b/runtime/src/tests.rs
@@ -560,43 +560,36 @@ mod pallets_interact {
 mod payout {
     use pallet_staking::EraPayout;
 
-    use crate::{Balance, Runtime, CENTS};
+    use crate::{Balance, Runtime};
 
     use super::new_test_ext;
 
-    /// Test that validators receive a cumulative reward that mimics the current emission of
-    /// $ZEN in the PoW Horizen blockchain for miners, which is a coinbase of 6.25 Zen for
-    /// each block every 2.5 minutes.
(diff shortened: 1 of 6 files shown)


runtime/src/tests.rs
```diff
@@ -560,43 +560,36 @@ mod pallets_interact {
...
 
-    use crate::{Balance, Runtime, CENTS};
+    use crate::{Balance, Runtime};
 
     use super::new_test_ext;
 
-    /// Test that validators receive a cumulative reward that mimics the current emission of
-    /// $ZEN in the PoW Horizen blockchain for miners, which is a coinbase of 6.25 Zen for
-    /// each block every 2.5 minutes.
+    #[ignore]
```
JackPiri, line 567: Maybe it's better to remove the `ignore`.

JackPiri (approved):
The PR is fine to me.
Thanks

runtime/src/tests.rs
DanieleDiBenedetto: Done thanks

la10736 (approved):
LGTM

PR #50, opened 2024-05-09
Node key file is now copied to specific directory

JackPiri:
The copied file is then automatically used by the node

This PR is motivated by this:
Currently, for secret-phrase, at first startup there's a key injection based on the mounted secret-phrase and from here on the node will use the keys and not the secret-phrase; hence it is totally fine to remove the secret-phrase file (from the host machine as well as from the container) without breaking anything; if the secret-phrase is kept and mounted at a second startup, the reinjection is pointless (doesn't occur at all or just rewrite the keys with identical value; no matter).
**Currently, for node-key file, it must always be provided at any startup as a mounted volume. The node itself targets the file through arg `--node-key-file`; with this PR the file gets copied in the proper location and no arg is passed to `nh-node`, because the copied file is automatically selected by the node itself; hence it is totally fine to remove the node-key file (from the host machine as well as from the container).**

**This PR has nothing to do with safe secrets handling (encryption) on container-side, it is just for allowing removal of host-side secrets file (which are still inspectable on container side).**

### DIFF
--- a/docker/resources/envs/.env.local
+++ b/docker/resources/envs/.env.local
@@ -1,5 +1,7 @@
 # RUST_LOG=debug
 
+NH_NODE_KEY_FILE="/data/config/node_key.dat"
+
 # Node config
 NH_CONF_NAME="RpcNode"
 NH_CONF_BASE_PATH="/data/node"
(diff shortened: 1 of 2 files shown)


lander86 (approved):
LGTM, thanks @JackPiri

drgora (approved):
LGTM

PR #51, opened 2024-05-10
Md/weights

la10736:
Introduce 

- Script to generate weights file
- a weight module in runtime for _all_ used pallets that need a weght and we can compute
- a default hw reference file (we don't use the subrate one)
- the support for define profile and featature in docker image (useful to create a docker to run benchmark on instance)
- default docker file now use production profile (maybe is better to use `testnet` till we are in testent?)

The scripts are wired to handle both docker and node executable... we'll improve them in the future

NOTE:
- Some pallets dosn't have an automatic benchmark support (babe, grandpa)
- Some pallets cannot benchmarked due we use some dependecies that are not thinked for production and dsn't provide a benchamking support (staking)
- pallet_session crash (:cry:) and need futter investigation
- pallet_im_online need a larger number max validator to run benchmark but this just make bigger the max storage size and nothing else


pallets/settlement-fflonk/src/mock.rs
```diff
@@ -50,6 +50,19 @@ pub mod on_proof_verified {
     }
 }
 
+pub struct MockWeightInfo;
+
+impl MockWeightInfo {
+    pub const REF_TIME: u64 = 42;
+    pub const PROOF_SIZE: u64 = 24;
```
DanieleDiBenedetto, line 57: How did you set these ?

DanieleDiBenedetto (approved):

JackPiri (approved):

PR #56, opened 2024-05-15
Add mac install instructions to readme, update test runner to be mac compatible

rushby:
Also added the passed tests count to the test result output.

### DIFF
--- a/e2e-tests/test_runner.sh
+++ b/e2e-tests/test_runner.sh
@@ -20,7 +20,7 @@ TXT_BIBLK="\033[90;1m"
 TXT_NORML="\033[0m"
 
 # Please do not exceed 64 chars for each test filename - including the .zndsl extension
-TEST_LIST=$(find -name "*.zndsl" | sort);
+IFS=$'\n' TEST_LIST=($(find . -name "*.zndsl" | sort))
 
 # The return value of each zombinet invocation is always equal to the
 # number of failed tests among those listed in each .zndsl.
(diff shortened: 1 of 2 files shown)


e2e-tests/test_runner.sh
```diff
@@ -107,14 +121,15 @@ echo -e "┌──────────────────────
...
+printf  "│ ${TXT_BIBLK} Failed tests:          ${TXT_BIRED} %3d ${TXT_NORML}                                          │\n" "${TOT_FAIL_TESTS}"
+
 if [ ${TOT_FAIL_TESTS} -ne 0 ]; then
     echo -e "├────────────────────────────────────────────────────────────────────────┤"
-    printf  "│ ${TXT_BIBLK} Failed tests:          ${TXT_BIRED} %3d ${TXT_NORML}                                          │\n" "${TOT_FAIL_TESTS}"
     for failed_test in "${FAILED_TESTS[@]}"; do
         printf "│     - %-64s │\n" "${failed_test}"
     done
-    echo -e "└────────────────────────────────────────────────────────────────────────┘"
-    exit 1
```
la10736, line 117: I guess that there's a little undesirable side effect by remove these lines: if some tests fails the script not fail.
That could be an issue both for the developer that cannot leverage on script output to understand in the some test failed and for the CI.

Ok, removing the footer duplication could be nice but we should define an exit value variable to recover the right behavior.

la10736 (changes requested):
Just recover the old behavior: if some e2e test fails `tests_runner.sh` should exit with error.

e2e-tests/test_runner.sh
rushby: Thanks both, i've added the exit status back in.

la10736: @drgora : @JackPiri already open a PR to fix it (I've raised a red flag yesterday when I saw that Daniele's PR fixed some format issues on my last PR.

drgora: Great for the other PR, thank you.

drgora (approved):
Looks good to me, thanks!

PR #57, opened 2024-05-20
Chain spec are built from file

JackPiri:


DanieleDiBenedetto (approved):
LGTM ! Please update readme as to clarify the process to build a node to sync to the official testnet, or one for local development purposes

node/src/chain_spec.rs
```diff
@@ -183,6 +183,11 @@ pub fn local_config() -> Result<ChainSpec, String> {
 }
 
 pub fn testnet_config() -> Result<ChainSpec, String> {
+    ChainSpec::from_json_bytes(&include_bytes!("../chain-specs/zkverify_testnet.json")[..])
+}
+
+#[allow(dead_code)]
+pub fn testnet_config_build() -> Result<ChainSpec, String> {
```
la10736, line 190: Why don't use this function for `test-dev` chain name?

la10736 (changes requested):
May is better to introduce a new chain spec `test-dev`

node/src/chain_spec.rs
JackPiri: Addressed as per your suggestion (just opted for `test_build` naming).

drgora (approved):
Looks good to me, thanks!

node/src/chain_spec.rs
```diff
@@ -186,7 +186,7 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
     ChainSpec::from_json_bytes(&include_bytes!("../chain-specs/zkverify_testnet.json")[..])
 }
 
-#[allow(dead_code)]
+// To be used when building new testnet chain-spec
```
la10736, line 189: Please use `///` to make this comment also a function documentation

la10736 (changes requested):
Please use `///` to generate documentation too

node/src/chain_spec.rs
JackPiri: Modified accordingly.

ptagl (approved):

la10736 (approved):
LGTM GREAT!

PR #61, opened 2024-05-21
Setting "shell: bash"

JackPiri:
Since the Orchestrator PR, the CI is silencing errors coming from following workflows:

```
CI-build-test.yml
CI-e2e-test.yml
CI-lint-format.yml
```

This wrong behavior is due to the critical steps being piped with `tee` command:

```
cargo test --lib --bins --all-features --no-fail-fast --release 2>&1 | tee unit_tests_output.txt
cargo test --test '*' --all-features --no-fail-fast --release 2>&1 | tee integration_tests_output.txt
yarn test 2>&1 | tee e2e_test_output.txt
cargo clippy --release -- --deny warnings 2>&1 | tee linting_output.txt
cargo fmt --check 2>&1 | tee formatting_output.txt
```

resulting in pipe exit code being equal to `tee` exit code (`0`, always success), hence suppressing possible failure exit code of the first command.

The solution is to set option `set -o pipefail`; in GitHubActions this can be achieved setting `shell: bash` for the target step ([reference](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference)).

### DIFF
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -43,11 +43,11 @@ jobs:
       - name: Build
         run: cargo build --release
       - name: Unit tests
-        run: |
-          cargo test --lib --bins --all-features --no-fail-fast --release 2>&1 | tee unit_tests_output.txt
+        shell: bash
+        run: cargo test --lib --bins --all-features --no-fail-fast --release 2>&1 | tee unit_tests_output.txt
       - name: Integration tests
-        run: |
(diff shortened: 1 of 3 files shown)


la10736 (approved):
LGTM. 

PR #56 should also fix the E2E tests bug that was here from a while.

drgora (approved):
This is great, thanks!

PR #64, opened 2024-05-30
Update runtime configuration

drgora:
This PR updates the runtime configuration to increase the maximum number of authorities and voters that elections can handle.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -131,7 +131,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // The version of the runtime specification. A full node will not attempt to use its native
     //   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
-    spec_version: 2_000,
+    spec_version: 2_001,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 1,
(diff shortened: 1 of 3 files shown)


runtime/src/lib.rs
```diff
@@ -394,11 +395,11 @@ parameter_types! {
     pub HistoryDepth: u32 = 30; // Number of eras to keep in history. Older eras cannot be claimed.
 }
 
-pub const MAX_AUTHORITIES: u32 = 10;
-pub const MAX_VOTERS: u32 = 10;
+pub const MAX_AUTHORITIES: u32 = 32;
```
DanieleDiBenedetto, line 398: Duplication here with MaxActiveValidators ?

DanieleDiBenedetto (approved):

runtime/src/lib.rs
drgora: Not quite, `MAX_AUTHORITIES` accounts for all validators, that is active + eligible, whereas `MaxActiveValidators` accounts only for the subset of active validators.
They are different things, so I'd rather keep them separated in the code, even if in the current configuration we use the same value for both.

runtime/src/lib.rs
```diff
@@ -131,7 +131,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // The version of the runtime specification. A full node will not attempt to use its native
     //   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
-    spec_version: 2_000,
+    spec_version: 2_100,
```
la10736, line 134: We should change the runtime `Cargo.toml` too

la10736 (changes requested):
Just change also cargo runtime version too

runtime/src/lib.rs
drgora: Done, thanks!

runtime/src/lib.rs
drgora: I have renamed `MAX_AUTHORITIES` and added some comments, I hope it's more intuitive now

la10736 (approved):
LGTM

ptagl (approved):

PR #65, opened 2024-06-04
More pipeline improve

ysibirski:
The following PR transition all the RUST workflows to run inside sc-ci-base docker container. Small changes and improvements were added along the lines as well. The PR also includes all the conditions like env.ACT to be able to run workflows locally without using cloud gh actions service. Run workflows locally functionality will be added in the nearest future


otoumas (approved):
LGTM

.github/workflows/CI-coverage.yml
```diff
@@ -1,65 +1,61 @@
...
-      matrix:
-        os: [ubuntu-latest]
-        # os: [ubuntu-latest, windows-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
-      - name: Checkout branch
+      - name: Git checkout
         uses: actions/checkout@v4
-      - name: Clean CI
```
JackPiri, line 21: I see `clean_ci.sh` script has been removed, that was useful for cleaning up some unused stuff inside GitHubActions runner; indeed now I am getting some errors like `No space left on device`.

.github/workflows/CI-build-test.yml
```diff
@@ -1,69 +1,65 @@
...
+        continue-on-error: true
         with:
           path: |
             ~/.cargo/bin/
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-build-test-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: ${{ runner.os }}-build-test-cargo-
```
JackPiri, line 32: I think `restore-keys` should be reintroduced, it was useful as a fallback in case an exact match on `key` was not found.
Please also apply to other workflows.

ci/entrypoint.sh
```diff
@@ -0,0 +1,90 @@
...
+  apt --no-install-recommends install -y nodejs
+  npm install -g yarn
+  echo -e "Nodejs environment was successfully deployed. Node.js version: $(node -v) | npm version: $(npm -v) | yarn version: $(yarn -v)\n"
+fi
+
+# System info
+rustup show
+if command -v java; then
+  echo -e "\n$(java --version)"
+fi
```
JackPiri, line 68: Very minor comment, I think this part can be revomed.

.github/workflows/CI-build-test.yml
```diff
@@ -1,69 +1,65 @@
...
 
-run-name: "Workflow performing CI steps: build, unit testing, integration testing"
+run-name: "Workflow CI/CD Steps: build, unit and integration testing"
 
 on:
   workflow_call:
 
 env:
-  RUST_BACKTRACE: 1
+  IMAGE_NAME: zencash/sc-ci-base
```
JackPiri, line 9: Very minor question, is it fine we are using this image named `zencash`? Do we want to duplicate it with different name?

.github/workflows/CI-build-test.yml
```diff
@@ -1,69 +1,65 @@
...
-      # required by nh-runtime
+          key: build-test-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Set docker env vars
         run: |
-            sudo apt-get install protobuf-compiler
-            rustup target add wasm32-unknown-unknown
-            rustup component add rust-src
-      - name: Check
-        run: cargo check --all-features --all-targets --release
```
JackPiri, line 42: It's not related to your PR (already present) but please take this chance for removing `--release` here (we are already using `--all-targets`)

.github/workflows/CI-build-test.yml
ysibirski: Done

ci/entrypoint.sh
ysibirski: This is not needed correct.

.github/workflows/CI-build-test.yml
ysibirski: This is a good question. We build these images under zencash workspace. I will need @lander86 and @cronicc input on it.

.github/workflows/CI-build-test.yml
ysibirski: Done

.github/workflows/CI-coverage.yml
JackPiri: Yes sure, it's on the `NH-core-ci_testing` repo.
https://github.com/HorizenLabs/NH-core-ci_testing/actions/runs/9366241226/job/25783220335

ysibirski: I will add cleanup.sh script back. I see now what is going on. :+1:

.github/workflows/CI-build-test.yml
ysibirski: @JackPiri those are public images with the open source content. Should be OK to use them.

ysibirski:
@JackPiri and @lander86 I have addressed all the comments. Please take a look again once you have time

JackPiri (approved):
The PR is fine to me.
Thanks

.github/workflows/CI-build-test.yml
lander86: As just mentioned by @ysibirski , it's ok to use those image.

lander86 (approved):
LGTM !

PR #67, opened 2024-06-06
Md/abstract_verifier

la10736:
Complete implementation and example

What is missed:

- E2E test for fflonk_verifier
- inclusion test in runtime


primitives/hp-verifiers/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+pub trait Verifier: 'static {
+    /// The proof format type accepted by the verifier
+    type Proof: Arg;
+    /// The public inputs format
+    type Pubs: Arg;
+    /// The verification key format
+    type Vk: VkArg;
+
+    /// The context used to generate the statements hash.
+    fn hash_context_data() -> &'static [u8];
```
DanieleDiBenedetto, line 59: This can be a constant perhaps ?

DanieleDiBenedetto (approved):

pallets/verifiers/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    fn statement_hash(ctx: &[u8], vk_hash: &H256, pubs: &[u8]) -> H256 {
+        let mut data_to_hash = ctx.to_vec();
+        data_to_hash.extend_from_slice(b"-");
+        data_to_hash.extend_from_slice(vk_hash.as_bytes());
+        data_to_hash.extend_from_slice(b"-");
+        data_to_hash.extend_from_slice(pubs);
+        H256(keccak_256(data_to_hash.as_slice()))
+    }
```
lgiussan, line 136: The way we encode the information here does not follow cryptographic best practices. In our application it is important that the hash uniquely identify a statement. However, both `ctx` and `pubs` do not have fixed sizes, and the encoding we are using makes it possible to define multiple statements that hash to the same value.
While it's true that "we" control `ctx` (which should make an exploit improbable/impossible), I fear that this is an issue which could be flagged in a future audit, since this is a well-known vulnerability (e.g. https://scsfg.io/hackers/abi-hash-collisions/).
To keep safe, we should use an encoding of the statement that prevents equivocation (e.g. by prepending `ctx` with its length)

primitives/hp-verifiers/src/lib.rs
la10736: No it's not... Function intraits cannot be declared as `const`

DanieleDiBenedetto (approved):

lgiussan (approved):
LGTM, great work!

PR #68, opened 2024-06-07
Update GRANDPA

drgora:
This PR does two things.
First it updates the GRANDPA configuration with safer parameters:
- `gossip_duration` is increased to account for bigger network latencies
- `voting_rule` is set to avoid finalizing blocks with higher probability of being reorganized

Second, it integrates the GRANDPA rpc, so that GRANDPA related functions are available on the RPC interface.

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -240,27 +240,46 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
     let enable_grandpa = !config.disable_grandpa;
     let prometheus_registry = config.prometheus_registry().cloned();
 
-    let rpc_extensions_builder = {
+    let (rpc_extensions_builder, shared_voter_state) = {
         let client = client.clone();
-        let pool = transaction_pool.clone();
+        let transaction_pool = transaction_pool.clone();
         let select_chain = select_chain.clone();
-        let babe = {
-            crate::rpc::BabeDeps {
-                keystore: keystore_container.keystore().clone(),
-                babe_worker_handle: babe_worker_handle.clone(),
-            }
-        };
 
-        Box::new(move |deny_unsafe, _| {
-            let deps = crate::rpc::FullDeps {
-                client: client.clone(),
-                pool: pool.clone(),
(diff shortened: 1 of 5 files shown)


node/src/service.rs
```diff
@@ -241,27 +240,46 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
...
+
+        // GRANDPA
+        let shared_voter_state_rpc = sc_consensus_grandpa::SharedVoterState::empty();
+        let shared_voter_state = shared_voter_state_rpc.clone();
+        let shared_authority_set = grandpa_link.shared_authority_set().clone();
+        let justification_stream = grandpa_link.justification_stream();
+        let finality_provider = sc_consensus_grandpa::FinalityProofProvider::new_for_service(
+            backend.clone(),
+            Some(shared_authority_set.clone()),
+        );
```
DanieleDiBenedetto, line 259: minor aesthetic note: can you bring this outside the code block declared at line 243 ?

node/src/service.rs
```diff
@@ -337,16 +355,19 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
 
         let grandpa_config = sc_consensus_grandpa::Config {
             // FIXME #1578 make this available through chainspec
-            gossip_duration: Duration::from_millis(333),
+            gossip_duration: Duration::from_millis(1000),
```
DanieleDiBenedetto, line 358: Is this affecting block finalization time as well ?

la10736, line 358: Where this cme from?

la10736 (approved):
LGTM

DanieleDiBenedetto (approved):

node/src/service.rs
drgora: It could. With this modification one GRANDPA round is potentially longer than before, and so finalization might be delayed accordingly.
Notice, we are talking about few seconds in the worst case anyway. For example, here is some interesting data on experimenting with a `2s` timeout versus `1s`.

PR #69, opened 2024-06-10
Added pallet for risc0 verifier

JackPiri:


pallets/settlement-risc0/src/benchmarking.rs
```diff
@@ -0,0 +1,36 @@
...
+    submit_proof {
+        let caller: T::AccountId = whitelisted_caller();
+    }: _(RawOrigin::Signed(caller), VALID_VK.into(), VALID_PROOF.to_vec().into(), VALID_PUBS.to_vec().into())
+}
+
+impl_benchmark_test_suite!(
+    SettlementRisc0Pallet,
+    crate::mock::new_test_ext(),
+    crate::mock::Test,
+);
```
la10736, line 36: That's quite semplicistic. This benchmark doesn't take care that the verify time is not costants but depends by program's cycle number that shoulb reflected in the proof size.

@DanieleDiBenedetto For now we can take this like it is bu maybe is better to add a ticket to handle it later.

pallets/settlement-risc0/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+            proof: Box<Vec<u8>>,
+            pubs: Box<Vec<u8>>,
+        ) -> DispatchResultWithPostInfo {
+            log::trace!("Submitting proof");
+            verify_proof::<T>(vk_u8, *proof, *pubs)
+                .map(Into::into)
+                .map_err(Into::into)
+        }
+    }
+
```
la10736, line 119: You can remove all `Box` here: take care of remove also all into inthe related calls

la10736 (changes requested):
Great work. Overall is OK... Just remove the unneeded `Box`'s in submit_proof extrinsic, add integration and E2E tests

runtime/src/lib.rs
```diff
@@ -340,6 +340,11 @@ impl pallet_settlement_zksync::Config for Runtime {
     type WeightInfo = weights::pallet_settlement_zksync::NHWeight<Runtime>;
 }
 
+impl pallet_settlement_risc0::Config for Runtime {
+    type OnProofVerified = Poe;
+    type WeightInfo = weights::pallet_settlement_risc0::NHWeight<Runtime>;
```
la10736, line 345: Maybe you neet to change it

runtime/src/tests.rs
```diff
@@ -192,6 +192,26 @@ fn pallet_zksync_availability() {
...
+#[test]
+fn pallet_risc0_availability() {
+    new_test_ext().execute_with(|| {
+        let dummy_origin = AccountId32::new([0; 32]);
+
+        let dummy_vk = [0u8; 32];
+        let dummy_proof = vec![];
+        let dummy_pubs = vec![];
+
+        assert!(SettlementRisc0Pallet::submit_proof(
```
la10736, line 204: Add your test here

la10736: And also the availability one

runtime/src/weights.rs
```diff
@@ -5,6 +5,7 @@ pub mod pallet_balances;
 pub mod pallet_im_online;
 pub mod pallet_poe;
```
la10736, line 6: You missed your module here

pallets/settlement-risc0/src/benchmarking.rs
DanieleDiBenedetto: Sure

DanieleDiBenedetto (approved):

runtime/src/lib.rs
JackPiri: Yes, modified together with runtime tests.

runtime/src/tests.rs
JackPiri: Added test for pallet availability and weight linking.

runtime/src/weights.rs
JackPiri: Yes, added together with runtime tests.

la10736 (approved):

PR #70, opened 2024-06-10
Allow manual workflow trigger

ysibirski:


lander86 (approved):

.github/workflows/CI-orchestrator.yml
```diff
@@ -3,10 +3,6 @@ name: Orchestrator
 on:
   pull_request:
     types: [opened]
-  push:
-    branches:
-      - main
-      - release/*
```
JackPiri, line 9: I think we cannot remove this part, especially if merge-queue process is still to be implemented/activated.
The problem is that in this way we are running CI just at PR opening (not at synchronized) and we are not running at push on `main` (e.g. merge commit). This can be dangerous because wrong commits can go to `main` branch without any notification.

I understand this has been done for fixing the situation _"Workflow does not exist or does not have a workflow_dispatch trigger in this branch."_ but I think we have to go for another solution; furthermore, I saw that in `NH-core-ci_testing` repo the `CI-orhestrator` job is able to run on non-`main` branches, so I think it's a matter of repo configuration.

JackPiri (changes requested):
I think we need to have an offline synchronization about this changes, please see my comment on `CI-orchestrator.yml`.

.github/workflows/CI-orchestrator.yml
lander86: Let's chat about the strategy in today's meeting @JackPiri

JackPiri: As agreed in meeting, let's proceed as it is.

JackPiri (approved):
The PR is fine to me.
Thanks

DanieleDiBenedetto:
Hold before merging plz, I will do it :)

ysibirski:
@DanieleDiBenedetto the latest commit should fix pipeline trigger by `on` events including tag releases

PR #71, opened 2024-06-11
Updated weights for 0.3.0 on reference machine

DanieleDiBenedetto:


lgiussan (commented):
The PR updates all the permissions of the weight files, making them executable (from `644` to `755`, i.e. from `-rw-r--r--` to `-rwxr-xr-x`). Is this something intended/needed?

DanieleDiBenedetto:
> The PR updates all the permissions of the weight files, making them executable (from `644` to `755`, i.e. from `-rw-r--r--` to `-rwxr-xr-x`). Is this something intended/needed?

Nope sorry, just fixed it

PR #72, opened 2024-06-11
Bump node, runtime and spec version to 0.3.0

la10736:


la10736:
I'm used a conservative approach ... Anyway I'll take a look to the documentation to check if we really need it

la10736:
> So no need to upgrade impl_version ?

No, impl version dosn't mater and not change on our side: name and version are related to different implementation of the same specification, I guess it's not our case

DanieleDiBenedetto:
> > So no need to upgrade impl_version ?
> 
> No, impl version dosn't mater and not change on our side: name and version are related to different implementation of the same specification, I guess it's not our case

Thanks for double checking !

lgiussan (approved):

PR #73, opened 2024-06-11
feat: Add multisig pallet

rdoria1:


runtime/src/lib.rs
```diff
@@ -329,6 +329,19 @@ impl pallet_sudo::Config for Runtime {
...
 }
 
+parameter_types! {
+    pub const MaxSignatories: u32 = 100;
+}
+impl pallet_multisig::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type RuntimeCall = RuntimeCall;
+    type Currency = Balances;
+    type DepositBase = ConstU128<EXISTENTIAL_DEPOSIT>;
```
rdoria1, line 339: To be discussed: 
DepositBase -> The base amount of currency needed to reserve for creating a multisig execution or to store a dispatch call for later. set as EXISTENTIAL_DEPOSIT for now

runtime/src/lib.rs
```diff
@@ -329,6 +329,19 @@ impl pallet_sudo::Config for Runtime {
...
+    type DepositFactor = ConstU128<0>;
```
rdoria1, line 340: To be discussed: 
DepositFactor -> The amount of currency needed per unit threshold when creating a multisig execution. Recall: The deposit to be made by the account that creates the multisig is: `threshold * DepositFactor + DepositBase` 
Set as 0 for now

runtime/src/weights/pallet_multisig.rs
```diff
@@ -0,0 +1,273 @@
+// Copyright 2024, Horizen Labs, Inc.
```
rdoria1, line 1: This file will be replaced once we run the benchmark in remote machine

runtime/src/weights/pallet_multisig.rs
```diff
@@ -0,0 +1,273 @@
...
+pub trait WeightInfo {
+    fn as_multi_threshold_1(z: u32, ) -> Weight;
+    fn as_multi_create(s: u32, z: u32, ) -> Weight;
+    fn as_multi_approve(s: u32, z: u32, ) -> Weight;
+    fn as_multi_complete(s: u32, z: u32, ) -> Weight;
+    fn approve_as_multi_create(s: u32, ) -> Weight;
+    fn approve_as_multi_approve(s: u32, ) -> Weight;
+    fn cancel_as_multi(s: u32, ) -> Weight;
+}
+
```
la10736, line 36: Ok that's a temp file but remove this useless trait from this file ... at least to shut up the warring :smile:

runtime/src/lib.rs
la10736: I prefer to define a new constant like `BASE_MULTISIG_DEPOSIT` to avoid undesiderable side effects.

la10736: You can use the same pattern used in polkadot and kusama:


define
```rust
parameter_types! {
	// One storage item; key size is 32; value is size 4+4+16+32 bytes = 56 bytes.
	pub const MultisigDepositBase: Balance = deposit(1, 88);
	// Additional storage item size of 32 bytes.
	pub const MultisigDepositFactor: Balance = deposit(0, 32);
	pub const MaxSignatories: u32 = 100;
}
```

Where the function `deposit` is defined in constant crate and seams to capture the storage cost in some sense:

```rust
	pub const fn deposit(items: u32, bytes: u32) -> Balance {
		items as Balance * 2_000 * CENTS + (bytes as Balance) * 100 * MILLICENTS
	}
```

runtime/src/tests.rs
```diff
@@ -23,12 +23,14 @@ use frame_support::{
         ExistenceRequirement, OnInitialize, WithdrawReasons,
     },
 };
+use frame_support::dispatch::{GetDispatchInfo};
```
la10736, line 26: Please remove unneeded curly braces

runtime/src/tests.rs
```diff
@@ -130,6 +138,156 @@ fn new_test_ext() -> sp_io::TestExternalities {
...
+        assert_eq!(existing_sudo_key, Some(account_ids[0].clone()));
+
+        // Setting the multisig account as the new sudo account
+        assert_ok!(Sudo::set_key(RuntimeOrigin::signed(account_ids[0].clone()), MultiAddress::Id(multi.clone())));
+
+        // Ensure the sudo key is set correctly
+        assert_eq!(Sudo::key(), Some(multi.clone()));
+
+        // Prepare a sudo call to change the sudo key again to a new account (account_ids[1])
+        let sudo_call = RuntimeCall::Sudo(SudoCall::set_key {
```
la10736, line 261: use `pallet_sudo::Call` and remove the unneeded indiretion here (you use it just once)

runtime/src/tests.rs
```diff
@@ -130,6 +138,156 @@ fn new_test_ext() -> sp_io::TestExternalities {
...
+        let account_ids: Vec<AccountId32> = testsfixtures::SAMPLE_USERS.iter().map(|user| AccountId32::from(user.raw_account)).collect();
+
+        let multi = Multisig::multi_account_id(&account_ids[..3], 2);
+
+        assert_ok!(Balances::transfer_allow_death(RuntimeOrigin::signed(account_ids[0].clone()), MultiAddress::Id(multi.clone()), 500));
+        assert_ok!(Balances::transfer_allow_death(RuntimeOrigin::signed(account_ids[1].clone()), MultiAddress::Id(multi.clone()), 500));
+        assert_ok!(Balances::transfer_allow_death(RuntimeOrigin::signed(account_ids[2].clone()), MultiAddress::Id(multi.clone()), 500));
+
+        // Balance for account 0 reduced by 500, after it was sent to the multisig
+        assert_eq!(Balances::free_balance(account_ids[0].clone()), testsfixtures::SAMPLE_USERS[0].starting_balance - 500);
```
la10736, line 157: You did a great job on writing these explorative tests that are useful to understand how the pallet works but here you're essentially testing substrate's code. I'm incline to trust substrate code and write just tests that trying to verify almost all possible the code that we wrote.

My idea is to have essentially the following tests:

1. `pallet_multisig_availability()`: just check you can use this pallet by call any kind of available extrinsic on this pallet - if this test compile you have a proof that you have had added this feature at your chain
2. `use_correct_weights::pallet_multisig()`: you already wrote 
3. `pallets_interact::sudo_multisig`: just move `test_set_sudo_to_multisig()` in `pallets_interact` module and change name; that's exactly a bussines case that we need to check and have a test that is here in place to attesting this case is useful.
4. (_Not sure about this test_) `pallet_multisig_configuration()`: here you should test that you configured the pallet correctly with some bussines cases.... the hard part here is what're the bussines cases maybe the deposit should be enought to cover the storage cost but what we are testing? that the constants are equals to some other constants?... Maybe that's really useless.

So maybe is better to remove this and the follow test... :cry:

la10736 (changes requested):
It's a great job but you need to fix
- format
- linting
- introcude dedicated constants for base deposit and deposit factor
- Review the testing police a little

Maybe would be nice to have an e2e test too: have a mutisig support is a business requirment so an E2E that attestate it seams reasonable.

runtime/src/lib.rs
rdoria1: Ok, after some research I think I understand more or less the logic in that computation (deposit method = the cost of storage based on the number of items (storage entries) and the size of those items in bytes)

I changed to these BUT the result are in wei format (or whoever is it called in Substrate) and based on what I saw how value are used in the implementation (existential deposit and tests ) it looks like we use whole numbers.

In order it be consistent I added an additional 

`pub const MultisigDepositBase: Balance = deposit(1, 88) / ACME;`
`pub const MultisigDepositFactor: Balance = deposit(0, 32) / ACME;`

given that `ACME = 1*10^18`

runtime/src/tests.rs
rdoria1: `pallet_multisig_availability` -> actually that's exactly my first test, I will keep that one as that one uses an extrinsic in the pallet

`use_correct_weights::pallet_multisig()` -> that one is done as you said

`pallets_interact::sudo_multisig` -> I'll move this one to pallets_interact as requested to a new module called sudo

I'll remove `cancel_multisig_returns_deposit` 

About `pallet_multisig_configuration` yeah I think we can skip it for now.

rdoria1:
> It's a great job but you need to fix
> 
> * format
> * linting
> * introcude dedicated constants for base deposit and deposit factor
> * Review the testing police a little
> 
> Maybe would be nice to have an e2e test too: have a mutisig support is a business requirment so an E2E that attestate it seams reasonable.

runtime/src/lib.rs
drgora: I understand the problem and your solution, but I'm not quite sure this is the best approach.
As it stands now, the deposit for multisig operations is almost dust, and that might not be what we want.
I would suggest instead to remove the dividers from the runtime constants, and adjust test balances to be aligned with real world cases. For example, you could multiply each test balance (for all the tests, not only the ones in this PR) by `ACME`?

runtime/src/tests.rs
```diff
@@ -723,6 +817,70 @@ mod pallets_interact {
...
+
+                // Check existing sudo key
+                let existing_sudo_key = Sudo::key();
+                assert_eq!(existing_sudo_key, Some(account_ids[0].clone()));
+
+                // Setting the multisig account as the new sudo account
+                assert_ok!(Sudo::set_key(
+                    RuntimeOrigin::signed(account_ids[0].clone()),
+                    MultiAddress::Id(multi.clone())
+                ));
```
drgora, line 848: Minuscule thing here. The following is more robust wrt the specific runtime configuration?
```suggestion
                // Setting the multisig account as the new sudo account
                assert_ok!(Sudo::set_key(
                    RuntimeOrigin::signed(existing_sudo_key.expect("No sudo key")),
                    MultiAddress::Id(multi.clone())
                ));
```

runtime/src/tests.rs
```diff
@@ -130,6 +137,83 @@ fn new_test_ext() -> sp_io::TestExternalities {
...
+
+        // Reserve balance was recovered, as this is just a safety measur until the tx is confirmed (the actual transfer comes from the multisig)
+        assert_eq!(
+            Balances::free_balance(account_ids[0].clone()),
+            testsfixtures::SAMPLE_USERS[0].starting_balance - 500
+        );
+        assert_eq!(Balances::reserved_balance(account_ids[0].clone()), 0);
+    });
+}
+
```
la10736, line 216: What I mean here is that a test like follow is enought and provide an assurance that the runtime include and configure multisign pallet without bother the reader with lot of logic to understand.

```suggestion
#[test]
fn pallet_multisig_availability() {
    new_test_ext().execute_with(|| {
        let issuer: AccountId32 = testsfixtures::SAMPLE_USERS[0].raw_account.into();
        let account_ids: Vec<_> = testsfixtures::SAMPLE_USERS
            .iter()
            .skip(1)
            .map(|u| u.raw_account.into())
            .collect();

        let call = Box::new(RuntimeCall::Balances(BalancesCall::transfer_allow_death {
            dest: MultiAddress::Id(issuer.clone()),
            value: 5000,
        }));

        assert_ok!(Multisig::as_multi(
            RuntimeOrigin::signed(issuer),
            2,
            account_ids,
            None,
            call,
            Weight::zero()
        ));
    })
}

```

runtime/src/tests.rs
```diff
@@ -130,6 +137,83 @@ fn new_test_ext() -> sp_io::TestExternalities {
...
 }
 
+fn call_transfer(dest: AccountId32, value: u128) -> Box<RuntimeCall> {
+    Box::new(RuntimeCall::Balances(BalancesCall::transfer_allow_death {
+        dest: MultiAddress::Id(dest),
+        value,
+    }))
+}
+#[test]
+fn test_multisig_send_and_return() {
```
drgora, line 147: I agree with @la10736 on this test testing the multisig pallet rather than our own runtime, so I would also be in favor of having something simpler just to check that you can compile a test that uses a multisig exstrinsic.
But if you do want to keep it, a small suggestion would be to replace hardcoded values with named variables; that might make the test more readable. For example, `500` could become something like `SEND_AMOUNT`, and `account_ids[0]` might become `depositor_id`.

drgora (commented):
Just a reminder to rebase this branch on `main` before merging the PR (I know it still takes some time for the weights).

runtime/src/tests.rs
DanieleDiBenedetto: Agreed ! Be aware of the fact that these pallets are already tested so no need to re-test the business logic here. It makes sense to add a test for our specific purpose (e.g. set sudo as multisig and check it works by making it execute a sudo call), and that can be a JS E2E test

runtime/src/lib.rs
la10736: My advice here is to remove `/ ACME` part and mutiply in `testfixtures.rs` file the `starting_balance` fields by `ACME`, leave just the last untouched... in the meantime you can remove the hardcoded `EXISTENTIAL_DEPOSIT` and import it from `crate::EXISTENTIAL_DEPOSIT`.

runtime/src/tests.rs
```diff
@@ -122,13 +123,42 @@ fn new_test_ext() -> sp_io::TestExternalities {
...
     let mut ext = sp_io::TestExternalities::from(t);
 
     ext.execute_with(|| System::set_block_number(1));
 
     // Return the test externalities
     ext
 }
+#[test]
+fn pallet_multisig_availability() {
+    new_test_ext().execute_with(|| {
```
la10736, line 141: THX!
Just move it closer to the others `pallet_*_availability()` tests.

runtime/src/lib.rs
rdoria1: following the mentioned approach

la10736 (approved):
LGTM

Really a great job!
Thx

drgora (approved):
Looks good to me, thanks!

PR #74, opened 2024-06-12
Telemetry update

JackPiri:
Since we would never generate a new chain spec file from scratch (apart from the case of a chain full regenesis), the associated interface has been removed. This was done also to avoid confusion, but the underlying code is kept as dead code (we can also remove it if we prefer).

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -30,7 +30,7 @@ const BOOTNODE_2_DNS: &str = "bootnode-tn-2.zkverify.io";
 const BOOTNODE_2_PEER_ID: &str = "12D3KooWEjVadU1YWyfDGvyRXPbCq2rXhzJtXaG4RxJZBkGE9Aug";
 
 // The URL for the telemetry server.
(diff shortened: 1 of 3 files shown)


node/src/chain_spec.rs
```diff
@@ -187,6 +187,7 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
 }
 
 /// To be used when building new testnet chain-spec
+#[allow(dead_code)]
```
la10736, line 190: i saw your point and I was the one that requested to leave it but now I prefer to remove it at all.

Anyway, feel free to leave it you prefer.

la10736 (approved):
LGTM

drgora (approved):
LGTM, thanks

node/src/chain_spec.rs
JackPiri: Yes probably even better to fully remove it. In case of need we could just navigate back in the repo history.

JackPiri: Oh sorry, I was late few seconds with my last push.
So right now we still have that dead code, no problem.

PR #76, opened 2024-06-14
Increased pubs max size, moved from constants to pallet configuration

JackPiri:
The actual values for `MAX_PROOF_SIZE` and `MAX_PUBS_SIZE` are defined at pallet level because they need to be used in:

- `runtime/src/lib.rs`
- `pallets/settlement-risc0/src/mock.rs`
- `pallets/settlement-risc0/src/tests.rs`

otherwise some code duplication would have been introduced.

They are defined as `u32` (instead of `usize`) because Substrate seems to lack `ConstUsize` macro.

### DIFF
--- a/pallets/settlement-risc0/src/lib.rs
+++ b/pallets/settlement-risc0/src/lib.rs
@@ -27,13 +27,15 @@ mod tests;
 mod weight;
 
 pub use weight::WeightInfo;
-pub const MAX_PROOF_SIZE: usize = 1000000;
-pub const MAX_PUBS_SIZE: usize = 32;
+pub const MAX_PROOF_SIZE: u32 = 1000000; // arbitrary length
+pub const MAX_PUBS_SIZE: u32 = 8 + 4 + 32 * 64; // 8: for bincode::serialize,
+                                                // 4: bytes for payload length,
+                                                // 32 * 64: sufficient multiple of 32 bytes
 
 #[frame_support::pallet]
 pub mod pallet {
-    use super::{WeightInfo, MAX_PROOF_SIZE, MAX_PUBS_SIZE};
-    use frame_support::{dispatch::DispatchResultWithPostInfo, ensure};
+    use super::WeightInfo;
+    use frame_support::{dispatch::DispatchResultWithPostInfo, ensure, pallet_prelude::Get};
     use frame_system::pallet_prelude::OriginFor;
     use hp_poe::OnProofVerified;
     use sp_core::H256;
@@ -75,6 +77,10 @@ pub mod pallet {
         type OnProofVerified: OnProofVerified;
         /// The weight definition for this pallet
         type WeightInfo: WeightInfo;
+        /// Maximum number of bytes contained in the proof (otherwise rejected)
+        type MaxProofSize: Get<u32>;
+        /// Maximum number of bytes contained in the public inputs (otherwise rejected)
+        type MaxPubsSize: Get<u32>;
     }
 
     pub fn verify_proof<T: Config>(
@@ -84,11 +90,11 @@ pub mod pallet {
     ) -> Result<(), Error<T>> {
         log::trace!("Checking size");
         ensure!(
-            (proof).len() <= MAX_PROOF_SIZE,
+            (proof).len() <= T::MaxProofSize::get().try_into().unwrap(),
             Error::<T>::InvalidProofSize
         );
         ensure!(
-            (pubs).len() <= MAX_PUBS_SIZE,
+            (pubs).len() <= T::MaxPubsSize::get().try_into().unwrap(),
             Error::<T>::InvalidPublicInputsSize
         );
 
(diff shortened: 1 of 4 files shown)


runtime/src/lib.rs
```diff
@@ -343,6 +343,8 @@ impl pallet_settlement_zksync::Config for Runtime {
 impl pallet_settlement_risc0::Config for Runtime {
     type OnProofVerified = Poe;
     type WeightInfo = weights::pallet_settlement_risc0::NHWeight<Runtime>;
+    type MaxProofSize = ConstU32<{ pallet_settlement_risc0::MAX_PROOF_SIZE }>;
+    type MaxPubsSize = ConstU32<{ pallet_settlement_risc0::MAX_PUBS_SIZE }>;
 }
```
la10736, line 349:
```suggestion
parameter_types! {
    pub const Risc0MaxProofSize: u32 = 1000000; // arbitrary length
    pub const Risc0MaxPubsSize: u32 = 8 + 4 + 32 * 64; // 8: for bincode::serialize,
                                                // 4: bytes for payload length,
                                                // 32 * 64: sufficient multiple of 32 bytes
}

impl pallet_settlement_risc0::Config for Runtime {
    type OnProofVerified = Poe;
    type WeightInfo = weights::pallet_settlement_risc0::NHWeight<Runtime>;
    type MaxProofSize = Risc0MaxProofSize;
    type MaxPubsSize = Risc0MaxPubsSize;
}
```

pallets/settlement-risc0/src/lib.rs
```diff
@@ -84,11 +90,11 @@ pub mod pallet {
...
         ensure!(
-            (proof).len() <= MAX_PROOF_SIZE,
+            (proof).len() <= T::MaxProofSize::get().try_into().unwrap(),
             Error::<T>::InvalidProofSize
         );
         ensure!(
-            (pubs).len() <= MAX_PUBS_SIZE,
+            (pubs).len() <= T::MaxPubsSize::get().try_into().unwrap(),
             Error::<T>::InvalidPublicInputsSize
         );
```
la10736, line 99:
```suggestion
        ensure!(
            (proof).len() <= T::MaxProofSize::get() as usize,
            Error::<T>::InvalidProofSize
        );
        ensure!(
            (pubs).len() <= T::MaxPubsSize::get() as usize,
            Error::<T>::InvalidPublicInputsSize
        );
```
From `u32` to `usize` you can use `as` directly

pallets/settlement-risc0/src/mock.rs
```diff
@@ -83,6 +84,8 @@ impl frame_system::Config for Test {
 impl crate::Config for Test {
     type OnProofVerified = OnProofVerifiedMock;
     type WeightInfo = MockWeightInfo;
+    type MaxProofSize = ConstU32<{ crate::MAX_PROOF_SIZE }>;
+    type MaxPubsSize = ConstU32<{ crate::MAX_PUBS_SIZE }>;
 }
```
la10736, line 90:
```suggestion
pub const MAX_PROOF_SIZE: u32 = 300000;
pub const MAX_PUBS_SIZE: u32 = 32;

impl crate::Config for Test {
    type OnProofVerified = OnProofVerifiedMock;
    type WeightInfo = MockWeightInfo;
    type MaxProofSize = ConstU32<{ MAX_PROOF_SIZE }>;
    type MaxPubsSize = ConstU32<{ MAX_PUBS_SIZE }>;
}
```

la10736 (changes requested):
I prefer to leave the runtime configuration in the runtime and not in the pallet.

Moreover some conversion can be simpler: `u32` can be always converted to `usize` without fails.

pallets/settlement-risc0/src/lib.rs
JackPiri: Modified as suggested.

pallets/settlement-risc0/src/mock.rs
JackPiri: Addressed.

runtime/src/lib.rs
JackPiri: Modified as suggested.

pallets/settlement-risc0/src/lib.rs
```diff
@@ -27,13 +27,15 @@ mod tests;
 mod weight;
 
 pub use weight::WeightInfo;
-pub const MAX_PROOF_SIZE: usize = 1000000;
-pub const MAX_PUBS_SIZE: usize = 32;
+pub const MAX_PROOF_SIZE: u32 = 1000000; // arbitrary length
+pub const MAX_PUBS_SIZE: u32 = 8 + 4 + 32 * 64; // 8: for bincode::serialize,
+                                                // 4: bytes for payload length,
+                                                // 32 * 64: sufficient multiple of 32 bytes
```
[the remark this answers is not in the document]
JackPiri: Addressed.

la10736 (approved):
LGTM

PR #78, opened 2024-06-17
feat: Add pallet_scheduler and pallet_preimage

rdoria1:


runtime/src/lib.rs
```diff
@@ -350,6 +353,79 @@ impl pallet_multisig::Config for Runtime {
...
+    }
+}
+impl pallet_scheduler::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type RuntimeOrigin = RuntimeOrigin;
+    type PalletsOrigin = OriginCaller;
+    type RuntimeCall = RuntimeCall;
+    type MaximumWeight = MaximumSchedulerWeight;
+    type ScheduleOrigin = EnsureRootOrSudo<Runtime>;
+
```
la10736, line 422: I think that we don't need this. `EnsureRoot` is enought.

We can just use a sudo account to schedule the action and that's will forward the ation to the scheduler pallet as it is Root. (I've tryed it)


```suggestion
    type ScheduleOrigin = EnsureRoot<AccountId>;
```

runtime/src/lib.rs
```diff
@@ -350,6 +353,79 @@ impl pallet_multisig::Config for Runtime {
...
+    }
+
+    #[cfg(feature = "runtime-benchmarks")]
+    fn try_successful_origin() -> Result<O, ()> {
+        pallet_sudo::Pallet::<T>::key()
+            .map(|key| O::from(RawOrigin::Signed(key)))
+            .or_else(|| Some(O::from(RawOrigin::Root))) // Fallback to root if no sudo key
+            .ok_or(())
+    }
+}
```
la10736, line 414: We can remove it at all

la10736 (changes requested):
Great: just 2 notes:

1. Maybe we don't need of `EnsureRootOrSudo` on `pallet-scheduler` and we can use just `EnsureRoot`
2. Can you try to write a test in `pallet_interact` module to check if `pallet_scheduler` use the configured `pallet_preimage`? hint: execute scheduler call and check preimge storage for the new entry

runtime/src/lib.rs
```diff
@@ -350,6 +353,79 @@ impl pallet_multisig::Config for Runtime {
...
+        frame_support::traits::LinearStoragePrice<
+            PreimageBaseDeposit,
+            PreimageByteDeposit,
+            Balance,
+        >,
+    >;
+}
+
+parameter_types! {
+    pub MaximumSchedulerWeight: Weight = BlockWeights::get().max_block;
```
DanieleDiBenedetto, line 380: Without knowing more about this, I would suggest to set it equal to Polkadot: 80% of the max block weight

runtime/src/lib.rs
```diff
@@ -301,9 +304,9 @@ impl pallet_balances::Config for Runtime {
     type WeightInfo = weights::pallet_balances::NHWeight<Runtime>;
     type FreezeIdentifier = ();
     type MaxFreezes = ();
-    type RuntimeHoldReason = ();
+    type RuntimeHoldReason = RuntimeHoldReason;
     type RuntimeFreezeReason = ();
-    type MaxHolds = ();
+    type MaxHolds = ConstU32<1>;
```
DanieleDiBenedetto, line 309: Let's increase this by a bit maybe

DanieleDiBenedetto (approved):
Overall LGTM, agreed with Michele's suggestions as well

runtime/src/lib.rs
```diff
@@ -350,6 +353,79 @@ impl pallet_multisig::Config for Runtime {
     type WeightInfo = weights::pallet_multisig::NHWeight<Runtime>;
 }
 
+parameter_types! {
+    pub const PreimageBaseDeposit: Balance = deposit(2, 64);
+    pub const PreimageByteDeposit: Balance = deposit(0, 1);
```
DanieleDiBenedetto, line 358: Actually, can we lower a bit these numbers atm ? The same goes for the corresponding parameters in the pallet_multisig implementation. Our faucet currently gives just 1 ACME and we want a lower entry barrier so we can let users battle test the multisig and preimage functionalities

DanieleDiBenedetto (changes requested):

runtime/src/lib.rs
rdoria1: ok the question is how low we want to go?

As a reference, at the moment those numbers translate into

`PreimageBaseDeposit = 40.064 ACME`
`PreimageByteDeposit = 0.001 ACME`

`MultisigDepositBase = 20.088 ACME`
`MultisigDepositFactor = 0.032 ACME`

If the faucet value is this low (1 ACME) I feel we need to divide a factor of 10 to those values,

so changing the deposit implementation from

```
 pub const fn deposit(items: u32, bytes: u32) -> Balance {
        items as Balance * 2_000 * CENTS + (bytes as Balance) * 100 * MILLICENTS
    }
```

to

```
pub const fn deposit(items: u32, bytes: u32) -> Balance {
        items as Balance * 200 * CENTS + (bytes as Balance) * 100 * MILLICENTS
    }
```

Would that make sense?

la10736: I haven't a strong opinion about that. I thought that's we should review all fee policies to have an organic approach. For now we can just follow a greedy approach to make simple to try it.

rdoria1: so if I understand correctly, you agree with my suggestion for now right? @la10736

DanieleDiBenedetto: Yes, let's make it low enough so users can try. I will open a ticket to tune them properly later. Thanks !

la10736 (approved):

la10736:
I've removed the test about scheduler preimage configuration that I suggested to add because it didn't test anything (counter example : remove Preimage configuration in scheduler and the test always pass).

Test this configuration is really hard and maybe we need really big call.... we'll came back on in later.

DanieleDiBenedetto (approved):

PR #85, opened 2024-06-24
Gp/ci e2e slack notification summary

JackPiri:
Slack notification was missing e2e tests summary.

Previously:

`Zombienet Tests: success`

Now:

`E2E Tests: Passed: 7, Failed: 0, Runtime: 360.87s`

### DIFF
--- a/.github/scripts/summarize_tests.sh
+++ b/.github/scripts/summarize_tests.sh
@@ -5,6 +5,7 @@ set -eEuo pipefail
 unit_test_file="./build-and-test-output/unit_tests_output.txt"
 integration_test_file="./build-and-test-output/integration_tests_output.txt"
 coverage_report_file="./coverage-output/coverage_report.json"
+e2e_test_file="./e2e-test-output/e2e_test_output.txt"
 
 # Initialize counters for unit tests
 unit_total_passed=0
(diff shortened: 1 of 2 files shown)


.github/scripts/summarize_tests.sh
```diff
@@ -5,6 +5,7 @@ set -eEuo pipefail
 unit_test_file="./build-and-test-output/unit_tests_output.txt"
 integration_test_file="./build-and-test-output/integration_tests_output.txt"
 coverage_report_file="./coverage-output/coverage_report.json"
+e2e_test_file="./e2e-test-output/e2e_test_output.txt"
```
otoumas, line 8: It's out of the scope of this PR, but maybe it would be better to have a single outputs directory and then create inside it all the reports. 
I think a good pattern would be:
```
unit_test_file="./outputs/unit_tests_report.txt"
integration_test_file="./outputs/integration_tests_report.txt"
coverage_report_file="./outputs/coverage_report.json"
e2e_test_file="./outputs/e2e_test_report.txt"
```

otoumas (approved):
LGTM, added a non blocking comment

.github/scripts/summarize_tests.sh
JackPiri: Ok makes sense, addressed inside last commit.

JackPiri: No unfortunately I am having some issues trying to quickly fix this; I'll addrss it in a later PR.
Thanks

DanieleDiBenedetto (approved):

PR #86, opened 2024-06-25
We cannot use the maximum crate value in benchmarks because the runtime is configured just for 16 and den't accept more.

la10736:
We cannot use the maximum crate value in benchmarks because the runtime is configured just for 16 and den't accept more.

Also add benchmarking tests that was be able to catch this issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5261,6 +5261,8 @@ dependencies = [
  "rstest_reuse",
  "scale-info",
  "sp-core",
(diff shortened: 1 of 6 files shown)


DanieleDiBenedetto (approved):

drgora (approved):
LGTM, thanks!

JackPiri (approved):
The PR is fine to me.
Thanks

PR #93, opened 2024-07-03
feat: Add script to create balances snapshot + restore balances

rdoria1:


scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -0,0 +1,86 @@
...
+    const balances = {};
+    const accounts = await api.query.system.account.keys();
+
+    for (const key of accounts) {
+        const account = key.args[0].toString();
+        if (DISCARDED_ACCOUNTS.includes(account)) continue;
+        const { data: { free: balance } } = await api.query.system.account(account);
+        balances[account] = balance.toString();
+    }
+    const sortedBalances = Object.entries(balances).sort((a, b) => (BigInt(b[1]) > BigInt(a[1]) ? 1 : -1));
```
DanieleDiBenedetto, line 28: Curiosity: why do you need to sort them ?

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -0,0 +1,86 @@
...
+    const balances = JSON.parse(fs.readFileSync(snapshotBalancesFile));
+    const custodyAccount = keyring.addFromUri(custodySeed);
+    let nonce = (await api.query.system.account(custodyAccount.address)).nonce.toNumber();
+
+    for (const [account, balance] of Object.entries(balances)) {
+        const transfer = api.tx.balances.transferAllowDeath(account, balance);
+        try {
+            const hash = await transfer.signAndSend(custodyAccount, { nonce });
+            console.log(`Transferred to ${account}, transaction hash: ${hash.toHex()}`);
+            nonce++;
```
DanieleDiBenedetto, line 49: Isn't the internal nonce of the custody account updated automatically at each signAndSend ?

DanieleDiBenedetto (approved):
LGTM ! Left a couple of comments but nothing blocking the merging.

Would it be possible to add a test with maybe a test snapshot and check that everything works properly ?

Moreover, It would be nice, as it might be useful in the future, to find a way to embed such balances in the chainspec

scripts/snapshot-restore-balance/snapshot-balances.js
rdoria1: It's not really needed. I did it to see which accounts were the whales and to make sure we had enough to distribute from the custody account. I can remove this if you prefer.

DanieleDiBenedetto: Maybe let's trigger it with an optional flag passed when invoking the script as it might widen the execution time if the snapshot starts to become very big

rdoria1: I think we can just remove it, we don't really need it

scripts/snapshot-restore-balance/snapshot-balances.js
rdoria1: looks like nop, if I remove the optional nonce all the transactions fail with `The transaction has too low priority to replace another transaction already in the pool.`

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -0,0 +1,81 @@
...
+    ]
+
+async function takeSnapshot(api, snapshotBalancesFile) {
+    const balances = {};
+    const accounts = await api.query.system.account.keys();
+
+    for (const key of accounts) {
+        const account = key.args[0].toString();
+        if (DISCARDED_ACCOUNTS.includes(account)) continue;
+        const { data: { free: balance } } = await api.query.system.account(account);
```
drgora, line 25: I might be missing some context, but I believe that we should also consider the `reserved` balance here.
For example, `reserved` balance should be transformed back into `free` balance if the operations that the balanced was reserved for are lost in the regenesis.

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -0,0 +1,81 @@
...
+    let nonce = (await api.query.system.account(custodyAccount.address)).nonce.toNumber();
+
+    for (const [account, balance] of Object.entries(balances)) {
+        const transfer = api.tx.balances.transferAllowDeath(account, balance);
+        try {
+            const hash = await transfer.signAndSend(custodyAccount, { nonce });
+            console.log(`Transferred to ${account}, transaction hash: ${hash.toHex()}`);
+            nonce++;
+            // Add a small delay between transactions to avoid priority issues
+            await new Promise(resolve => setTimeout(resolve, 500));
```
drgora, line 46: This is interesting; what kind of priority issues are you referring to? Is it related to the ordering of transactions?

I'm asking because, in general, this kind of timeouts tend to be very fragile in terms of functionality and sub-optimal in terms of performance. It would be great to find a formal solution to the issue you are mitigating with the timeout.

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -0,0 +1,81 @@
...
+        }
+        await takeSnapshot(api, arg1);
+    } else if (command === 'restore') {
+        if (!arg1 || !arg2) {
+            console.error('Please provide the custody seed and snapshot balances file name.');
+            process.exit(1);
+        }
+        await restoreBalances(api, keyring, arg2, arg1);
+    } else {
+        console.error('Unknown command. Use "snapshot" or "restore".');
```
drgora, line 73: It would be very helpful to have a very brief usage example, to clarify what parameters should be passed without having to look at the code.

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    fs.writeFileSync(snapshotBalancesFile, JSON.stringify(balances, null, 2));
+
+    console.log('Snapshot saved.');
+}
+
+// Restore balances from the snapshot
+async function restoreBalances(api, keyring, snapshotBalancesFile, custodySeed) {
+    const balances = JSON.parse(fs.readFileSync(snapshotBalancesFile));
+    const custodyAccount = keyring.addFromUri(custodySeed);
```
drgora, line 36: Very small suggestion here. What about adding a sanity check on the balance of the `custodyAccount` (also accounting for the expected fees) before start sending the extrinsics?
We might catch immediately a condition that would break the restore process in the middle, otherwise.

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -0,0 +1,81 @@
...
+    let nonce = (await api.query.system.account(custodyAccount.address)).nonce.toNumber();
+
+    for (const [account, balance] of Object.entries(balances)) {
+        const transfer = api.tx.balances.transferAllowDeath(account, balance);
```
drgora, line 40: If the existential deposit is increased between the dump and restore, the custody account could be removed before the loop completes. Probably this risk can be negligible if we restore balances in reverse balance order, starting from the one with the least amount. 

Otherwise we might also switch to `transferKeepAlive`, and handle possible errors accordingly.

drgora (changes requested):
Thank you for the PR. I left few comments, most of which are not really blocking.
Still, I believe we should really consider `reserved` balances in the dump and restore process, before merging.

scripts/snapshot-restore-balance/snapshot-balances.js
rdoria1: Adding examples in the error message

scripts/snapshot-restore-balance/snapshot-balances.js
rdoria1: @drgora 
Actually my bad, it's just I had a "priority issue" but that was actually fixed by bumping the nonce, I did not need this timeout on top of that.

That being said, isn't there going to be any issue about submitting too many transactions within a same block? Or do they just sit in the mempool for next block(s)? 

I will remove the timeout for now

scripts/snapshot-restore-balance/snapshot-balances.js
rdoria1: I see what you mean but I guess given the amount of accounts this is very unlikely to happen right ? anyway I will update the takeSnapshot function so that we save in reverse order and that order is respected in the restore one

scripts/snapshot-restore-balance/snapshot-balances.js
drgora: Ok great, I'm glad to see this solved.

Also, you are right; even if you send more transactions than can fit in a single block, Substrate will handle everything as you would expect: they will stay in the mempool until they find a spot in a block,

drgora (approved):
Thank you for addressing the feedback, it looks great to me now!

PR #94, opened 2024-07-04
Add scripts for reference machine benchmarking:

cronicc:
* ./scripts/reference_machine_run_all.sh
  * Sets CPU frequency and io profiles to simulate an AWS c7a.2xlarge instance with EBS io2 8000 IOPS storage.
  * Script has to be run as root/with sudo.
  * Script detects if it's running on the benchmark machine and can only be run there.
  * Machine CPU frequency settings will be restored to defaults after execution and temporary storage is cleaned up.
  * Script sets up a simulated block device stored in memory (/dev/shm) with performance throttled, this ensures consistent performance between runs.
* ./scripts/run_all_benchmarks.sh and ./scripts/bench.sh
  * Scripts were modified to allow setting --base-path argument for 'nh-node benchmark'.
  * support was added to build the docker image automatically with '--features=runtime-benchmarks'.
  * Docker Compose is used to set I/O limits, 'docker run' CLI switches to set I/O limits are not working due to a bug in docker upstream.
  * Two compose yml files were added, one for running on developers local machine, one for running on the benchmark machine with I/O throttling.


AndrewCoal737 (approved):
This looks great!

lander86 (approved):
Looks good !

scripts/run_all_benchmarks.sh
```diff
@@ -19,44 +19,59 @@
...
-if [ -z "${USE_DOCKER_IMAGE}" ]; 
-then
-echo "[+] Compiling nh-node benchmarks..."
-cargo build \
-  --locked \
-  --features=runtime-benchmarks \
-  --bin nh-node
+if [ "${USE_DOCKER}" = "false" ]; then
+  echo "[+] Compiling nh-node benchmarks..."
+  cargo build \
```
JackPiri, line 47: I know it was already this way before current PR, but I was wondering if we want to build with profile `production` here, since we are using `PROFILE=production` in the Docker case.
Something like:
  cargo build \
    --locked \
    --features=runtime-benchmarks \
    --profile production \
    --bin nh-node

  # The executable to use.
  NH_NODE="${PROJECT_ROOT}/target/release/nh-node"

JackPiri (commented):
The PR is fine to me.
I left just a comment for the building in the non-Docker scenario, please have a look at it, and if proper then address it.
Thanks

PR #95, opened 2024-07-10
Gp/tag alignement

JackPiri:
Following new tagging criteria.
Also some minor fixes for `ci/setup_env.sh` script are included.


.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -3,7 +3,7 @@ name: Tag-orchestrator
 on:
   push:
     tags:
-      - '[0-9].[0-9].[0-9]*'
+      - '[0-9]+.[0-9]+.[0-9]+-[0-9]+.[0-9]+.[0-9]+*'
```
ysibirski, line 6: This regex enforces the following format for all the tags 
`DIGIT.DIGIT.DIGIT-DIGIT.DIGIT-DIGIT` that means nothing like D.D.D will match anymore. 

If you want to support both cases try to use the following regex and see if GH actions will accept the syntax. `\` in the regex to escape `.` since it is an actual DOT vs any character from regex syntax 

```^[0-9]+\.[0-9]+\.[0-9]+(-[0-9]+\.[0-9]+\.[0-9]+)?$```

File: ci/docker.sh [hunk 2]
```diff
@@ -37,32 +37,34 @@ if [ -z "${docker_hub_username:-}" ]; then
...
+  docker build --build-arg PROFILE=production -f "${docker_file_path}" -t "${docker_hub_org}/${docker_image_build_name}:${docker_tag_full}" .
 
   # Publishing to DockerHub
   log_info "=== Publishing Docker image(s) on Docker Hub ==="
   echo "${docker_hub_token}" | docker login -u "${docker_hub_username}" --password-stdin
 
   # Docker image(s) tags for PROD vs DEV release
   if [ "${prod_release}" = "true" ]; then
-    publish_tags=("${docker_tag}" "latest")
+    docker_tag_node=$(cut -d'-' -f1 <<< $docker_tag_full)
```
ysibirski, line 56: The idea is to publish 2 tags if it comes in the format of 1.1.1-2.2.2? So we want to publish 1.1.1-2.2.2 as well as 1.1.1?

ci/docker.sh
```diff
@@ -37,32 +37,34 @@ if [ -z "${docker_hub_username:-}" ]; then
...
+    publish_tags=("${docker_tag_full}" "${docker_tag_node}" "latest")
   elif [ "${dev_release}" = "true" ]; then
-    publish_tags=("${docker_tag}" "dev")
+        docker_tag_node=$(cut -d'-' -f1 <<< $docker_tag_full)-$(cut -d'-' -f3- <<< $docker_tag_full)
```
ysibirski, line 59: These cut commands will lead to the following value if TAG is 1.1.1-2.2.2 for example that is being introduced push by tags regex

`docker_tag_node=1.1.1-` which I think not what we want?

If tag is 1.1.1 then `docker_tag_node=1.1.1-2.2.2` 
It seems like the same is happening for PROD_RELEASE? Why not to use the same approach then? 

docker_full_tag=github_ref_name (= 1.1.1-2.2.2)

```
docker_tag_node="$(cut -d '-' -f1 <<< "${docker_tag_full}")"   (= 1.1.1)
```

ci/setup_env.sh
```diff
@@ -10,9 +10,9 @@ export COMMON_FILE_LOCATION='ci/common.sh'
 workdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
 github_tag="${GITHUB_REF_NAME:-}"
 release_branch="${RELEASE_BRANCH:-release}"
-prod_release_regex='^[0-9]+\.[0-9]+\.[0-9]+$'
-dev_release_regex='^[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+){1}$'
-test_release_regex='^[0-9]+\.[0-9]+\.[0-9]+-[a-zA-Z0-9]+$'
+prod_release_regex='^[0-9]+\.[0-9]+\.[0-9]+\-[0-9]+\.[0-9]+\.[0-9]+$'
```
ysibirski, line 13: Again, same here this regex will enforce 1.1.1-2.2.2 format and will not match 1.1.1 for example

ci/setup_env.sh
```diff
@@ -10,9 +10,9 @@ export COMMON_FILE_LOCATION='ci/common.sh'
...
+dev_release_regex='^[0-9]+\.[0-9]+\.[0-9]+\-[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+){1}$'
```
ysibirski, line 14: Same here

ci/setup_env.sh
```diff
@@ -10,9 +10,9 @@ export COMMON_FILE_LOCATION='ci/common.sh'
...
+test_release_regex='^[0-9]+\.[0-9]+\.[0-9]+\-[0-9]+\.[0-9]+\.[0-9]+-[a-zA-Z0-9]+$'
```
ysibirski, line 15: Same here

ci/setup_env.sh
```diff
@@ -64,15 +64,9 @@ log_info "Release branch(es) is(are): ${release_branch}/*"
 log_info "Github tag is: ${github_tag}"
 
 # Checking if it is a release build
-if git branch -r --contains "${github_tag}" | grep -xqE ". origin\/${release_branch}/[^/]+$"; then
+if git branch -r --contains "${github_tag}" | grep -xqE ". origin\/${release_branch}/${github_tag}"; then
```
ysibirski, line 67: The new desired approach here github tag deriving from `release/github_tag` branch? Therefore we consider release ONLY if tag = 3.3.3 for example and it MUST derive from `release/3.3.3` branch?

If this is the desired approach then everything looks good

ci/setup_env.sh
```diff
@@ -85,20 +79,14 @@ if git branch -r --contains "${github_tag}" | grep -xqE ". origin\/${release_bra
...
-        export PROD_RELEASE="true"
-      elif [[ "${github_tag}" =~ ${dev_release_regex} ]]; then
-        export DEV_RELEASE="true"
-      elif [[ "${github_tag}" =~ ${test_release_regex} ]] && ! [[ "${github_tag}" =~ -rc ]]; then
-        export TEST_RELEASE="true"
-      else
-        log_warn "WARNING: GitHub tag: ${github_tag} is in the wrong format for PRODUCTION, DEVELOPMENT or TEST release. Expecting the following format for the release: PRODUCTION = 'd.d.d' | DEVELOPMENT = 'd.d.d-rc[0-9]' | TEST = 'd.d.d-*'. The build is not going to be released ..."
-        export IS_A_RELEASE="false"
-      fi
+    if [[ "${github_tag}" =~ ${prod_release_regex} ]]; then
```
ysibirski, line 82: You do not need `release_name` var anymore then

ci/setup_env.sh
```diff
@@ -85,20 +79,14 @@ if git branch -r --contains "${github_tag}" | grep -xqE ". origin\/${release_bra
...
+      export PROD_RELEASE="true"
+    elif [[ "${github_tag}" =~ ${dev_release_regex} ]]; then
+      export DEV_RELEASE="true"
+    elif [[ "${github_tag}" =~ ${test_release_regex} ]] && ! [[ "${github_tag}" =~ -rc ]]; then
+      export TEST_RELEASE="true"
     else
-      log_warn "WARNING: GitHub tag = ${github_tag} does NOT match GitHub release branch name = ${release_name}. The build is not going to be released ..."
+      log_warn "WARNING: GitHub tag: ${github_tag} is in the wrong format for PRODUCTION, DEVELOPMENT or TEST release. Expecting the following format for the release: PRODUCTION = 'd.d.d' | DEVELOPMENT = 'd.d.d-rc[0-9]' | TEST = 'd.d.d-*'. The build is not going to be released ..."
```
ysibirski, line 89: Message needs to be adjusted because of the regex changes

ysibirski (changes requested):
See some comments

.github/workflows/CI-tag-orchestrator.yml
JackPiri: Yes that's correct, on the GitHub tags side there will be nomore `X.Y.Z` but only `X.Y.Z-A.B.C`; slighlty different for Docker tags.

ci/docker.sh
JackPiri: The answers to your questions are both yes.
Modified as per your suggestion.

ci/docker.sh
JackPiri: Modified as per your suggestion.

ci/setup_env.sh
JackPiri: Yes that's correct, on the GitHub tags side there will be nomore `X.Y.Z` but only `X.Y.Z-A.B.C`; slighlty different for Docker tags.

ci/setup_env.sh
JackPiri: See comment above for GitHub tagging new rule.

ci/setup_env.sh
JackPiri: Yes definitely, good catch.
Modified.

ci/setup_env.sh
JackPiri: Agree, removed the line.

ci/setup_env.sh
JackPiri: Yes, just to make a couple examples more:

- tag `3.3.3-3.3.3`, branch `release/3.3.3-3.3.3`,
- tag `3.3.3-3.3.3-rc1`, branch `release/3.3.3-3.3.3-rc1`,
- tag `3.3.3-3.3.3-temp`, branch `release/3.3.3-3.3.3-temp`.

This modification comes mainly from the following non-working scenario we had before:

- tag `3.3.3-rc1` (created at commit sha `abcd`), branch `release/3.3.3-rc1`,
- tag `3.3.3` (created at commit sha `abcd`), branch `release/3.3.3`,
- falling into error case `WARNING: More than 1 GitHub '${release_branch}/*' branch contains current GitHub tag: ${github_tag}. The build is not going to be released ...`.

ysibirski (approved):
LGTM

DanieleDiBenedetto (approved):

PR #96, opened 2024-07-12
Renaming NH to ZKV

drgora:
This is a first attempt at updating all the legacy New Horizen (and NH acronyms) references to zkVerify (ZKV).

**Notice**: the spec name of the runtime has not been changed, to preserve compatibility with the live testnet.


docker/dockerfiles/binary_injected.Dockerfile
```diff
@@ -8,22 +8,22 @@ ARG BUILD_DATE
...
 ARG VENDOR="Horizen Labs"
 
-LABEL io.hl.image.authors=${AUTHORS} \
-	io.hl.image.vendor="${VENDOR}" \
-	io.hl.image.revision="${VCS_REF}" \
-	io.hl.image.title="${IMAGE_NAME}" \
-	io.hl.image.created="${BUILD_DATE}" \
-	io.hl.image.documentation="${DOC_URL}" \
-	io.hl.image.description="${DESCRIPTION}" \
-	io.hl.image.source="https://github.com/HorizenLabs/zkVerify/blob/${VCS_REF}/docker/dockerfiles/binary_injected.Dockerfile"
```
drgora, line 26: I'm uncertain on these renamings `hl` -> `zkv`. What do you guys think?

la10736: Here I prefer hl instead. It's the vendor and not the product.

drgora: Ok makes sense for these variables, thanks.

I'd leave the renaming of the docker files though, since that is indeed the product.

runtime/src/lib.rs
```diff
@@ -132,8 +132,8 @@ pub mod opaque {
 // https://docs.substrate.io/main-docs/build/upgrade#runtime-versioning
 #[sp_version::runtime_version]
 pub const VERSION: RuntimeVersion = RuntimeVersion {
-    spec_name: create_runtime_str!("nh-node"),
```
la10736, line 135: Ok, we should understand if we want to re-genesis our chain. @DanieleDiBenedetto has told me that maybe we want.... maybe is better to decide what we want to do ASAP :smile:

docker/scripts/entrypoint.sh
```diff
@@ -6,23 +6,23 @@
 # - preparation before the node start (example keys injection)
 # - launch of the actual node
 # 
-# Environment variables should generally be in the form `NH_*`
```
JackPiri, line 9: Being the renaming from `NH_/NH_CONF` to `ZKV_/ZKV_CONF` a breaking change, let's make sure we align the `compose` projects accordingly.

pallets/proof_of_existence/src/weight.rs
```diff
@@ -22,7 +22,7 @@
 //! WASM-EXECUTION: `Compiled`, CHAIN: `Some("dev")`, DB CACHE: `1024`
 
 // Executed Command:
-// /home/mdamico/devel/NH-core/target/production/nh-node
```
JackPiri, line 25: This can be opinionated, so evaluate if actually to be considered.
Regardless of file autogeneration (and probably being this overwritten in the next future), since all weigths file has been modified within the PR I would also adapt:

- `the New Horizen node` to `the zkVerify node`,
- `NH-core/node` to `zkVerify/node`,
- `bin/nh-node` to `bin/zkv-node`.

This comment applies to all weight files.

JackPiri:
Not directly related to this PR, but since we are doing some renaming/cleaning, I stumbled upon `verifiers/groth16/src/resources.rs` header and I think we want to change `// Copyright 2024, The Horizen Foundation` to `// Copyright 2024, Horizen Labs, Inc.`.

JackPiri:
I am quite sure in `.github/workflows/CI-rustdoc.yml` we want to modify this `url=nh_node/index.html` to `url=zkv_node/index.html`, but please double check.

JackPiri (changes requested):
I've left some comments, not big stuff but I think it's better to address/read them before approving.
Thanks

runtime/src/lib.rs
DanieleDiBenedetto: Let's proceed without that for the moment, so we unlock other refactoring activities. When we will have a feedback we could open another PR.

pallets/proof_of_existence/src/weight.rs
DanieleDiBenedetto: I think that:
- The first comment is addressed by the change in `node/hl-deploy-weight-template.hbs`
- The second one is due to the fact that whoever run the benchmark still had the old "NH-core" name (didn't re-clone the repo when we changed the name)
- Same goes for the third one

docker/scripts/entrypoint.sh
DanieleDiBenedetto: Will merge this one and start adjusting also the other repos :)

DanieleDiBenedetto (approved):

JackPiri (approved):

PR #98, opened 2024-07-17
Gp/ci remove check

JackPiri:
This PR is for removing CI steps that can be considered almost useless as of now.

Removing `cargo check` is almost a no-brainer, since:

- it takes approx 16 minutes when cache is not available,
- it takes approx 3 minutes when cache is available,
- it is followed by `cargo build`.

Removing `cargo test --test '*'` (integration tests) is opinionated, since:

- it takes approx 2 minutes when cache is not available,
- it takes approx 1 minute when cache is available,
- it actually executes **NO TESTS**,
- can be kept for when actual integration tests will be introduced.

### DIFF
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -47,9 +47,6 @@ jobs:
           echo "USER_ID=$(id -u)" >> $GITHUB_ENV
           echo "GRP_ID=$(id -g)" >> $GITHUB_ENV
 
-      - name: Cargo check
-        run: ${{ env.DOCKER_COMPOSE_CMD }} cargo check --all-features --all-targets
-
       - name: Cargo build
         run: ${{ env.DOCKER_COMPOSE_CMD }} cargo build --release
 


DanieleDiBenedetto (approved):
LGTM !

.github/workflows/CI-build-test.yml
```diff
@@ -47,28 +47,20 @@ jobs:
...
         run: ${{ env.DOCKER_COMPOSE_CMD }} cargo build --release
 
       - name: Cargo unit tests
         shell: bash
         run: ${{ env.DOCKER_COMPOSE_CMD }} cargo test --lib --bins --all-features --no-fail-fast --release 2>&1 | tee unit_tests_output.txt
 
-      - name: Cargo integration tests
-        shell: bash
-        run: ${{ env.DOCKER_COMPOSE_CMD }} cargo test --test '*' --all-features --no-fail-fast --release 2>&1 | tee integration_tests_output.txt
-
```
la10736, line 63: In the relay chain implementation that I'm doing I've integrated some integration tests from polkadot... ok maybe we can live without them but I prefer to recover it.

la10736 (changes requested):
Please recover integration tests

.github/workflows/CI-build-test.yml
JackPiri: Ok reverted, integration tests are now back again.
Force pushing.

la10736 (approved):
LGTM

PR #99, opened 2024-07-19
Add open governance

rdoria1:


la10736 (approved):
That's a huge pile of numbers a tweaks ...

LGTM

runtime/src/governance/origins.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    #[pallet::pallet]
+    pub struct Pallet<T>(_);
+
+    #[derive(PartialEq, Eq, Clone, MaxEncodedLen, Encode, Decode, TypeInfo, RuntimeDebug)]
+    #[pallet::origin]
+    pub enum Origin {
+        /// Origin able to cancel slashes and manage minimum commission.
+        StakingAdmin,
+        /// Origin for spending up to $10,000,000 DOT from the treasury as well as generally
```
drgora, line 36: It looks like these comments are not aligned with the actual values below, could you please double check and update in case?

drgora (approved):
I left few minor comments, but it looks good to me. Thanks!

runtime/src/governance/origins.rs
rdoria1: values are fine, just changed DOT to ACME and put a TBD for the treasury value

PR #100, opened 2024-07-19
Gp/bench var size risc0

JackPiri:
Consider the following table for a generic guest program (the program whose execution you want to prove/verify, in risc0):

| Number of cycles (2^) | Proof size (bytes) |
| - | - |
| 12 | 215538 |
| 13 | 215538 |
| 14 | 238578 |
| 15 | 238578 |
| 16 | 238578 |
| 17 | 238578 |
| 18 | 250290 |
| 19 | 262514 |
| 20 | 275762 |
| 21 | 514256 |
| 22 | 789934 |
| 23 | 1353002 |
| 24 | 2455714 |

Two things to note:

1. in the verifier implementation a collapse is implemented for aggregating proofs backed by different number of cycles (considering power of 2) but coming with same size: `CYCLE_2_POW_FROM_12_TO_13`, `CYCLE_2_POW_FROM_14_TO_17`, ..., `CYCLE_2_POW_FROM_24_TO_24`,
2. running benchmarks, it is _almost_ confirmed that proof verification time is dependent on proof size.


la10736 (approved):
In general I approve it.

Anyway I noted that the execution time is linear in the proof size. So, we can simplify all things a lot:

We should just benchmarking three end-to-end and cases:
- proof size 2^12 `W0`
- proof size 2^24 `Wmax`
- proof size 2^12 with key hash `W0h`

Now if we call the size of the proof for 2^12 `S0`, `Smax` the one for 2^24 and `S` the size of the give proof we can apply the follow formula:

```
W = (S - S0) * K + Wbase
```
where:
- `K=(Wmax-W0)/(Smax - S0)`
- `Wbase = W0` in the case where no hash is provided or `Wbase = W0h` otherwise

lgiussan (approved):
This PR is a clear improvement over the previous state, where proof size was not taken into account for determining extrinsic weight, so I would suggest to merge it. However, at the moment, the benchmarks only cover the following scenario:
- [`InnerReceipt`](https://docs.rs/risc0-zkvm/latest/risc0_zkvm/enum.InnerReceipt.html) of type `Composite`. It could also be `Succinct`, and `Groth16`.
- `Poseidon2` hash function. It could also be `sha256`.
- Default [`segment_limit_po2`](https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.ExecutorEnvBuilder.html#method.segment_limit_po2) (i.e. 20). It could be anything between [`risc0_zkp::MIN_CYCLES_PO2`](https://docs.rs/risc0-zkp/1.0.3/x86_64-unknown-linux-gnu/risc0_zkp/constant.MIN_CYCLES_PO2.html) (i.e. 13), and [`risc0_zkp::MAX_CYCLES_PO2`](https://docs.rs/risc0-zkp/1.0.3/x86_64-unknown-linux-gnu/risc0_zkp/constant.MAX_CYCLES_PO2.html) (i.e. 24)

At the moment, since the proof is treated as an opaque object, it would be difficult to distinguish all the possible variations. In a future PR, we will have to find the best way to address this issue.

PR #101, opened 2024-07-24
feat: SS58 prefix change + rebalance script change

rdoria1:


runtime/src/lib.rs
```diff
@@ -211,7 +211,7 @@ parameter_types! {
 
     pub BlockLength: frame_system::limits::BlockLength = frame_system::limits::BlockLength
         ::max_with_normal_ratio(5 * 1024 * 1024, NORMAL_DISPATCH_RATIO);
-    pub const SS58Prefix: u8 = 42;
```
drgora, line 214: I believe it would be good to have a short comment saying why this value was chosen, i.e. what prefix it translates to in SS58

drgora (commented):
Thank you for the PR.

As I was testing, I noticed that the Talisman wallet does not consider the custom prefix read from the endpoint RPC, and it keeps using the default 42 instead. It looks like they rely on info from this repository instead: https://github.com/TalismanSociety/chaindata.

I guess that we want to also update Talisman repo before moving on with this change, otherwise we would have a confusing scenario where Talisman addresses do not match PolkadotJS' (and probably also Polkascan's?).

drgora (approved):
This PR is fine by me. I'd just like to reiterate that we should make sure we don't create confusion with Talisman wallets before merging.

runtime/src/lib.rs
```diff
@@ -211,7 +211,9 @@ parameter_types! {
...
+
+    // ASCII for ZKV
```
JackPiri, line 215: Maybe a bit pedantic but it's not immediate to get this explanation: I would rewrite `// ASCII for 'Z'+'K'+'V'`.
But just my taste, feel free to ignore.

node/src/chain_spec.rs
```diff
@@ -161,6 +162,7 @@ pub fn local_config() -> Result<ChainSpec, String> {
         let mut props = Properties::new();
         props.insert("tokenSymbol".into(), "ACME".into());
         props.insert("tokenDecimals".into(), 18.into());
+        props.insert("ss58Format".into(), zkv_runtime::SS58Prefix::get().into());
```
JackPiri, line 165: Please insert `props.insert("ss58Format".into(), zkv_runtime::SS58Prefix::get().into());` also in `testnet_config_build` (line 232/233).
I know it's dead code, but if possible let's keep it aligned.

JackPiri (approved):
Left just a couple of minor optional comments.
The PR is fine to me.
Thanks

PR #110, opened 2024-08-06
Fix regression on synch

la10736:
In order to avoid regressions we cannot change the old `enum` positions otherwise old calls will change


drgora (approved):
LGTM, thanks!

PR #111, opened 2024-08-06
Moved the data directory renaming before keys handling

JackPiri:
There are cases in which the keys handling results in `zkv_*` directory creation, hence the renaming would be skipped.
Also a fix for handling properly `ZKV_CONF_BASE_PATH` (instead of hard-coded `/data/node/` is provided.

Example:
Bootnodes come with ZKV_NODE_KEY_FILE evaluated and the actual file present.
This code
```
# Node-key handling
if [[ (-n "${ZKV_CONF_BASE_PATH:-}") && (-n "${ZKV_CONF_CHAIN:-}") && (-f "${ZKV_NODE_KEY_FILE}") ]]; then
...
mkdir -p "${destination}"
  echo "Copying node key file"
  cp "${ZKV_NODE_KEY_FILE}" "${destination}/secret_ed25519"
fi
```
results in the `zkv_*` folder being created before `nh_*` being renamed.
The subsequent check `[ ! -e "$dest_chain_dir" ]` fails and no renaming is performed.
The final symptom is the done resyncing from genesis.

### DIFF
--- a/docker/scripts/entrypoint.sh
+++ b/docker/scripts/entrypoint.sh
@@ -84,6 +84,18 @@ while IFS='=' read -r -d '' var_name var_value; do
   fi
 done < <(env -0)
 
+if [ -n "${ZKV_CONF_BASE_PATH}" ]; then
+  BASE_CHAINS="${ZKV_CONF_BASE_PATH}/chains"
+
+  for chain in local testnet ; do
+    source_chain_dir="${BASE_CHAINS}/nh_${chain}";
+    dest_chain_dir="${BASE_CHAINS}/zkv_${chain}";
+    [ -d "$source_chain_dir" ] && [ ! -e "$dest_chain_dir" ] && \
+      echo "Move ${source_chain_dir} to ${dest_chain_dir}" && \
+      mv "${source_chain_dir}" "${dest_chain_dir}"
+  done
+fi
+
 # Keys handling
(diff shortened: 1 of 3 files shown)


docker/scripts/entrypoint.sh
```diff
@@ -84,6 +84,18 @@ while IFS='=' read -r -d '' var_name var_value; do
...
 
+if [ -n "${ZKV_CONF_BASE_PATH}" ]; then
+  BASE_CHAINS="${ZKV_CONF_BASE_PATH}/chains"
+
+  for chain in local testnet ; do
+    source_chain_dir="${BASE_CHAINS}/nh_${chain}";
+    dest_chain_dir="${BASE_CHAINS}/zkv_${chain}";
+    [ -d "$source_chain_dir" ] && [ ! -e "$dest_chain_dir" ] && \
+      echo "Move ${source_chain_dir} to ${dest_chain_dir}" && \
+      mv "${source_chain_dir}" "${dest_chain_dir}"
```
DanieleDiBenedetto, line 95: Since we are changing this, should we use the `ls -s` command ?

la10736 (approved):

DanieleDiBenedetto (approved):

docker/scripts/entrypoint.sh
JackPiri: Mhm well I think there's no need, we just wanna make sure we have migrated/moved the data, not interested in which data actually.

PR #112, opened 2024-08-06
Treasury and Bounties Pallets

rdoria1:
**Definition and values**

**Treasury Pallet params**

`PalletId` -> Just an ID for the pallet, where the accound ID for the pallet is derived from. I chose zk/trsry
the format is always <SOME_NAME>/trsry but SOME_NAME is only limited to 2 chars, so I picked zk
`Currency`-> It specifies which type will be used to handle currency-related operations within the Treasury pallet, Set to Balances pallet as the currency handler for the Treasury pallet.
`ApproveOrigin` -> Define who can approve treasury proposals. In this case, it's either the root account or a designated Treasurer custom origin.
`RejectOrigin` -> Defines who can reject treasury proposals. In this case, it's either the root account or a designated Treasurer custom origin.
`OnSlash` -> It determines what happens to funds when they are slashed (taken as a penalty) from validators or other participants in the network. Set to Treasury pallet. meaning that  slashed funds will be redirected to the Treasury itself.
`ProposalBond` -> It is a deposit that must be paid when submitting a spending proposal to the treasury. This deposit serves as a deterrent against spam or frivolous proposals. The amount of the bond is typically a percentage of the proposed spend.  -> It is set to 5 %.  If someone proposes a spend of 100 tokens, they would need to deposit 5 tokens as a bond.
`ProposalBondMinimum` -> sets the minimum amount for a proposal bond, regardless of the proposal's value. -> Set to 20 ACMEs
`ProposalBondMaximum` -> sets the maximum amount for a proposal bond, capping it even for very large proposals. -> Set to 1000 ACMEs
`SpendPeriod` -> It is the length of time (in blocks) between treasury "spending" events. It determines how often the treasury can disburse funds for approved proposals. -> Currently set to 6 days
`Burn` -> Sets the percentage of unused funds that should be burned at the end of each spend period  -> 1% in this case
`SpendFunds` -> Defines a mechanism to spend funds from the treasury. -> currently set to pallet bounties meaning that creating and funding bounties is one of the main mechanisms for spending Treasury resources. Bounties pallet has other params but I won't put them here, let me know if you need them as well.
`MaxApprovals` -> sets the maximum number of spending proposals that can be approved in the Treasury at any given time. This parameter prevents an excessive number of approved but unfunded proposals from accumulating in the system. -> Currently set a 100
`SpendOrigi`n -> It defines the origin (or permission) required to spend funds directly from the treasury without going through the proposal process. -> Currently set to a specific custom origin TreasurySpender
`AssetKind` -> Defines the type of assets that the Treasury can handle. -> This was set with some cross chain stuff, but I changed it to empty as we don't need to handle asset types in this context. Valid config from other chains I checked in similar situation.
`Beneficiary` -> It specifies the type that represents the recipient of funds from the Treasury -> Was set with multichain params but I set it to AccounId meaning that Treasury payouts will only go to any valid account in zk verify chain.
`Paymaster` -> Defines how payments are made, implemented involving cross-chain transfers but I changed to  -> Indicates how payments are made from treasury pallet -> Was set to some multi chain stuff, changed so that when treasury makes to do a payment it will draw funds from specific treasury account that holds all the treasury funds.
`PayoutPeriod` -> Defines the duration for which a beneficiary can claim an approved payout from the Treasury. After a proposal is approved, the beneficiary has this period to claim their payout. -> currently set at 30 days

**Bounties Pallet params**

`BountyDepositBase` -> sets the base deposit required to create a bounty. Ensures a minimum stake to create a bounty, discouraging spam-> Set to 1 ACME
`BountyDepositPayoutDelay` -> Defines the delay period between a bounty being successfully completed and the payout being made. Allows time for any disputes or issues to be raised before payout. -> set to 8 days
`BountyUpdatePeriod` -> Sets the period after which a bounty status will be updated if not already updated. Ensures bounties don't remain stagnant indefinitely. -> Set to 90 days
`MaximumReasonLength` -> Defines the maximum length of the reason (description) for a bounty. Limits the storage space used by bounty descriptions. -> Set to 16,384 characters
Curators are individuals or entities responsible for managing and overseeing specific bounties within the system.
`CuratorDepositMultiplier` -> Sets the curator's deposit as a percentage of the bounty value. Ensures curators have a stake in the bounty's success. -> Set to 50%
`CuratorDepositMin` -> Sets the minimum deposit a curator must put down. Ensures a minimum stake for curators, even for small bounties. -> Set to 10 ACMEs
`CuratorDepositMax` -> Sets the maximum deposit a curator must put down. Caps the curator's risk for very large bounties. -> 200 ACMEs
`DataDepositPerByte` -> Sets the deposit required per byte of data in the bounty. Discourages unnecessarily large data attachments to bounties.
`ChildBountyManager` -> No child bounties as it's currently empty
`BountyValueMinimum` -> Sets the minimum value for a bounty to be created. Ensures bounties are of significant value to be worth processing. -> Set to 10 ACMEs


runtime/src/lib.rs
```diff
@@ -399,6 +400,71 @@ impl pallet_scheduler::Config for Runtime {
...
+    type SpendFunds = Bounties;
+    type SpendOrigin = TreasurySpender;
+    type AssetKind = ();
+    type Beneficiary = AccountId;
+    type BeneficiaryLookup = IdentityLookup<Self::Beneficiary>;
+    type Paymaster = PayFromAccount<Balances, ZKVerifyTreasuryAccount>;
+    type BalanceConverter = UnityAssetBalanceConversion;
+    type PayoutPeriod = PayoutSpendPeriod;
+    #[cfg(feature = "runtime-benchmarks")]
+    type BenchmarkHelper = ();
```
DanieleDiBenedetto, line 439: What is this used for ? I see that in Polkadot runtime is set

DanieleDiBenedetto (approved):
LGTM from code perspective. From parameters perspective I suggest to double check with John. I will provide some thoughts too

Regarding `AssetKind`, `Beneficiary` and `Paymaster` they were set to "multi-chain" stuff because Polkadot allows to support multiple asset types, as long as they are listed in the AssetHub system parachain (more info [here](https://wiki.polkadot.network/docs/learn-polkadot-opengov-treasury#multi-asset-treasury-support)). True that we are not interested in it atm, but maybe something we could consider supporting in the future.

runtime/src/lib.rs
DanieleDiBenedetto: Ah maybe it's related to the fact that Treasury supports spending assets from different parachains, so they need to simulate these. Not needed for now in our case I assume

runtime/src/lib.rs
```diff
@@ -399,6 +400,71 @@ impl pallet_scheduler::Config for Runtime {
...
+impl pallet_bounties::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type BountyDepositBase = BountyDepositBase;
+    type BountyDepositPayoutDelay = BountyDepositPayoutDelay;
+    type BountyUpdatePeriod = BountyUpdatePeriod;
+    type CuratorDepositMultiplier = CuratorDepositMultiplier;
+    type CuratorDepositMin = CuratorDepositMin;
+    type CuratorDepositMax = CuratorDepositMax;
+    type BountyValueMinimum = BountyValueMinimum;
+    type ChildBountyManager = ();
```
DanieleDiBenedetto, line 462: I'm afraid we might need also the ChildBounties pallet. Bounties are usually used to access relevant portion of Treasury funds and allocate them to some activities to be specified later via child bounties. Bounties are an easier way to access Treasury Funds when compared to OpenGov referenda. More info [here](https://wiki.polkadot.network/docs/learn-polkadot-opengov-treasury#bounties).

DanieleDiBenedetto (changes requested):

runtime/src/lib.rs
```diff
@@ -399,6 +400,71 @@ impl pallet_scheduler::Config for Runtime {
     type Preimages = Preimage;
 }
 
+parameter_types! {
+    pub const TreasuryPalletId: PalletId = PalletId(*b"zk/trsry");
+    pub const ProposalBond: Permill = Permill::from_percent(5);
+    pub const ProposalBondMinimum: Balance = 2000 * CENTS;
+    pub const ProposalBondMaximum: Balance = THOUSANDS;
+    pub const SpendPeriod: BlockNumber = 6 * DAYS;
+    pub const Burn: Permill = Permill::from_percent(1);
```
drgora, line 409: I'm not necessarily asking for a change here, but I'd like to draw attention to this.

Do we really want to commit to spending Treasury funds promptly? I'm wondering whether it might be better to set this to `0`, at least initially, until we have established practices for handling these funds.

drgora (changes requested):
Thank you, this PR looks mostly good to me!

One thing I believe we should do before merging, is updating the `pallet_staking` config so that `Slash` and `RewardRemainder` are assigned `Treasury`.
After this, it would also be great to have tests that check these connections.
For example, checking that Staking's slashes go to treasury might be a small extension of one test that we have already, `offences::notifies_staking`.

runtime/src/lib.rs
rdoria1: added

runtime/src/lib.rs
rdoria1: This is correct the `TreasuryArguments` defined in the default implementation is designed to support multichain support, which is why this is empty () in ours (confirmed in implementation on another chain that is single support like ours)

runtime/src/lib.rs
rdoria1: Ok will check this with John

rdoria1: @drgora checked with John, so

- Changed burn to 0 as requested (will change that 1 or something else later but not for now) 
- No changed SpendPeriod: His worry was that treasury would pay no matter what but that is not case, every 6 days fund will be paid only for approved proposals and if sufficient funds are available.

drgora: This is exactly what I had in mind, thanks!

DanieleDiBenedetto (approved):
LGTM !

drgora (approved):
Thank you, this looks good to me.
We can add tests to check the interaction between staking and treasury in a follow up PR.

PR #115, opened 2024-08-08
Fix git revision on node version (it was `unknow`)

la10736:
In build layer use the .gi folder to compute rev and set SUBSTRATE_CLI_GIT_COMMIT_HASH


JackPiri (approved):

drgora (approved):
LGTM

PR #116, opened 2024-08-08
feat: Add utility pallet

rdoria1:
Tried a few local batch calls from local node in Polkadot JS, works fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5637,6 +5637,23 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


la10736 (dismissed):
In general it's ok but some tests are missed:

- the weight configuration
- inclusion

drgora (approved):
I agree on @la10736's comment, but once that is done, this is ok for me.

lgiussan (approved):
LGTM

PR #118, opened 2024-08-09
Pallet vesting

rdoria1:


drgora (approved):

runtime/src/lib.rs
```diff
@@ -292,6 +292,23 @@ impl pallet_utility::Config for Runtime {
     type WeightInfo = weights::pallet_utility::ZKVWeight<Runtime>;
 }
 
+parameter_types! {
+    pub const MinVestedTransfer: Balance = 100 * CENTS;
+    pub UnvestedFundsAllowedWithdrawReasons: WithdrawReasons =
+        WithdrawReasons::except(WithdrawReasons::TRANSFER | WithdrawReasons::RESERVE);
+}
```
lgiussan, line 299: What's the rationale for this choice? My understanding is that this leaves open the possibility to withdraw unvested funds for the following three reasons: TRANSACTION_PAYMENT, FEE, and 'TIP`. Just wanted to double-check that we are ok with this.

rdoria1: that's correct. It means those 3 options are the only valid reasons to withdraw the funds.

The reason is really to follow the default polkadot implementation. But to be more specific on why they decided to do that this is my cut

Blocking TRANSFER and RESERVE for unvested funds makes sense because these actions could potentially allow a user to bypass the vesting schedule:

Where the other 3 actions, are generally allowed because they are necessary for participating in the chain’s operations. For example, even unvested funds might need to be used to pay for transaction fees or to provide small tips for validators to get transactions included. Allowing these operations ensures that users can still interact with the chain, even if their funds are not yet fully vested.

lgiussan: Makes sense, thanks!

lgiussan (approved):

PR #123, opened 2024-08-29
Implement caching dependency crate

la10736:
Here we implement a utility crate to cache native dependecy. This PR primiry fixing an annoing bug that trigger the build and linking of mainchain crate every times even if nothing change. A side effect of this issue is that vscode often stuck.

Clean start expected behavior (no envs and no deps folder)

- first build build everything: also rocksdb and ultraplonk
- second build native and mainchain and linking. In the meantime generate cache in `deps` and environment variables in `~/.cargo/config.toml`
- third build native and mainchain and linking (because the cache and environment variables are changed)
- forth build do nothing

After that even if you issue `cargo clean`  the next build will compile rocksdb and ultraplonk really fast.

*ATTENTION*:
Try to compile this PR on a previous sorce folder could cause some linking issues. Delete `deps` folder, remove env variables from `<CARGO_HOME>/config.toml` and run `cargo clean` reset all.


utils/native-cache/src/helpers.rs
```diff
@@ -0,0 +1,29 @@
...
+
+/// Identify if the current path is a static or dynamic lib
+pub fn is_dyn_or_static_lib(p: impl AsRef<Path>) -> bool {
+    let p = p.as_ref();
+    if let Some((name, ext)) = p
+        .file_name()
+        .and_then(OsStr::to_str)
+        .and_then(|n| p.extension().map(|e| (n, e)))
+    {
+        (ext == "a" || ext == "so") && name.starts_with("lib")
```
drgora, line 20: Not that it matters much as we are not using shared libraries, but still:
```suggestion
        (ext == "a" || ext == "so" || ext == "dylib") && name.starts_with("lib")
```

drgora (approved):
Looks good to me, thanks!

utils/native-cache/src/helpers.rs
la10736: THX

JackPiri (approved):
The PR is fine to me.
Thanks

PR #124, opened 2024-08-29
Don't save full attestation history

DanieleDiBenedetto:
This PR implements the following:

- A new MaxStorageProofs parameter has been added to the "Poe" pallet. We will keep in storage at most that number of attestations and the associated proofs' digests. () can be passed to keep the current behaviour (all attestations are saved)
- When publishing an attestation from root origin the storage isn't actually filled with it. These behaviour has been changed as to keep uniformity with the rest
- Benches have been adapted accordingly, however timings might have increased dramatically. We should reason more on this.

Any advice/suggestion is greatly welcomed ! :)


pallets/proof_of_existence/src/tests.rs
```diff
@@ -62,17 +63,18 @@ pub static HASHES: [[u8; 32]; 5] = [
...
         assert!(Poe::publish_attestation(RuntimeOrigin::root()).is_ok());
-        assert_attestation_evt(0, H256::default());
+        assert_attestation_evt(0, H256(Keccak256::hash(H256::default().as_bytes()).into()));
     })
 }
 
 #[test]
 fn root_publish_two_attestations() {
     new_test_ext().execute_with(|| {
+        let default_empty_att = H256(Keccak256::hash(H256::default().as_bytes()).into());
```
la10736, line 73: maybe extract a `default_empty_att()` function will result little bit nicer.

pallets/proof_of_existence/src/tests.rs
```diff
@@ -123,6 +125,63 @@ fn correct_root() {
...
+        let max_attestations = crate::mock::MAX_STORAGE_ATTESTATIONS.saturated_into::<u64>();
+
+        for i in 0..=max_attestations * 2 {
+            // Publish proofs and attestation
+            let pida = H256(HASHES[0]);
+            Poe::on_proof_verified(pida);
+            let pidb = H256(HASHES[1]);
+            Poe::on_proof_verified(pidb);
+
+            assert!(Poe::publish_attestation(RuntimeOrigin::root()).is_ok());
```
la10736, line 140: You can use `assert_ok!` here instead

pallets/proof_of_existence/src/lib.rs
```diff
@@ -164,6 +181,23 @@ pub mod pallet {
                 || (values > 0 && deadline.unwrap_or_default())
         }
 
+        fn prune_old_attestations_if_needed() {
+            let attestations = Values::<T>::iter_keys()
+                .map(|(k1, _)| k1)
+                .collect::<BTreeSet<_>>();
```
la10736, line 187: There's a big issue here: your are reading ALL proof storage in order to find all attestations index. A mitigation could be use `iter_key_prefix` that returns just an element for each attestation and decode de id by use `prefix()` call... I but that's not a solution.

Maybe you need to introduce a new storage value for the last removed id.

pallets/proof_of_existence/src/lib.rs
```diff
@@ -164,6 +181,23 @@ pub mod pallet {
...
+            let att_len = attestations.len().saturated_into::<usize>();
+            let max_len = T::MaxStorageAttestations::get().saturated_into::<usize>();
```
la10736, line 189:
```suggestion
            let att_len = attestations.len();
```

pallets/proof_of_existence/src/benchmarking.rs
```diff
@@ -13,23 +13,68 @@
...
+            }
+        }
+
+        for id in offset..max_attestations {
+            for _ in 0..T::MinProofsForPublishing::get() * 2 {
+                test_hash[0] += 1;
+                assert!(Values::<T>::try_get(id as u64, H256::from_slice(&test_hash)).is_ok());
+            }
+        }
+    }
```
la10736, line 75: You don't need to fill all values but from a benchmark point of view you must fill all attestations that you should remove with prune call and the one you'll publish.


```suggestion
    #[benchmark]
    fn publish_attestation() {
        // Setup code
        // benchmark taking into account situations where attestations left to be cleaned in storage
        // are bigger than one. shouldn't happen very often so we consider only 0.01% of the
        // maximum supported ones.
        let max_attestations = T::MaxStorageAttestations::get();
        let offset = max_attestations.div_ceil(OFFSET_FACTOR);
        for id in 0..max_attestations + offset {
            if id < offset {
                // benchmark with double the minimum number of elements for publishing an attestation
                let mut hash: [u8; 32] = [0; 32];
                for _ in 0..T::MinProofsForPublishing::get() * 2 {
                    hash[0] += 1;
                    Values::<T>::insert(id as u64, H256::from_slice(&hash), ());
                }
            } else {
                Values::<T>::insert(id as u64, H256::default(), ());
            }
        }

        #[extrinsic_call]
        publish_attestation(RawOrigin::Root);

        assert_eq!(Values::<T>::iter_keys().count(), max_attestations as usize);
    }
```

la10736 (changes requested):
You need to find a way to avoid the read of all storage data when you prune old attestation. Moreover you should update the weight module in the pallet crate because you changed the behaviour and so the default `()` implementation is changed.

pallets/proof_of_existence/src/lib.rs
```diff
@@ -164,6 +185,19 @@ pub mod pallet {
...
+            let oldest_attestation_id = Self::oldest_attestation();
+            let max_attestations = T::MaxStorageAttestations::get().saturated_into::<u64>();
+
+            if newest_attestation_id - oldest_attestation_id > (max_attestations - 1) {
+                for id in oldest_attestation_id..newest_attestation_id - (max_attestations - 1) {
+                    let _ = Values::<T>::clear_prefix(id, u32::MAX, None);
+                }
+                OldestAttestation::<T>::set(newest_attestation_id - (max_attestations - 1));
+            }
+        }
```
la10736, line 199: Be carful on use `max_attestations - 1` When the value is `0` just works for chance. I think you need some tests about `max_attestations = 0` behaviour

pallets/proof_of_existence/src/benchmarking.rs
```diff
@@ -13,23 +13,71 @@
...
 
-benchmarks! {
-    publish_attestation {
-        // benchmark with double the minimum number of elements for publishing an attestation
         let mut hash: [u8; 32] = [0; 32];
-        for h in 0 .. T::MinProofsForPublishing::get() * 2 {
-            hash[0] += 1;
-            Values::<T>::insert(0, H256::from_slice(&hash), ());
+
+        for id in 0..max_attestations + offset {
```
la10736, line 43: What's about to define `let attestations = max_attestations + offset` and make your intent more clear?

pallets/proof_of_existence/src/benchmarking.rs
```diff
@@ -13,23 +13,71 @@
...
+        let mut test_hash: [u8; 32] = [0; 32];
+        for id in 0..offset + 1 {
+            for _ in 0..T::MinProofsForPublishing::get() * 2 {
+                test_hash[0] += 1;
+                assert!(Values::<T>::try_get(id, H256::from_slice(&test_hash)).is_err());
+            }
+        }
+
+        for id in offset + 1..max_attestations + 1 {
+            assert!(Values::<T>::try_get(id, H256::default()).is_ok());
```
la10736, line 76:
```suggestion
        for id in offset + 1..offset + max_attestations {
```

pallets/proof_of_existence/src/benchmarking.rs
```diff
@@ -13,23 +13,71 @@
...
+
+    const OFFSET_FACTOR: u64 = 10000;
+    #[benchmark]
+    fn publish_attestation() {
+        // Setup code
+        // benchmark taking into account situations where attestations left to be cleaned in storage
+        // are bigger than one. shouldn't happen very often so we consider only 0.01% of the
+        // maximum supported ones.
+        let max_attestations = T::MaxStorageAttestations::get().saturated_into::<u64>();
+        let offset = max_attestations
```
la10736, line 37: I think you'r overestimate here: in this case you're considering the case where you're pruning (10 + 1) full attested tree with 32 leafs for each case. As far as I can understand you can have just one attestation to prune (32 proofs in the wroste case)

pallets/proof_of_existence/src/benchmarking.rs
```diff
@@ -13,23 +13,71 @@
...
-    publish_attestation {
-        // benchmark with double the minimum number of elements for publishing an attestation
         let mut hash: [u8; 32] = [0; 32];
-        for h in 0 .. T::MinProofsForPublishing::get() * 2 {
-            hash[0] += 1;
-            Values::<T>::insert(0, H256::from_slice(&hash), ());
+
+        for id in 0..max_attestations + offset {
+            if id < offset + 1 {
+                // benchmark with double the minimum number of elements for publishing an attestation
```
la10736, line 45: You should also fill the last one (the one that you'll publish)
```suggestion
            if id < offset + 1 || id == max_attestations + offset - 1 {
```

la10736 (changes requested):
You should consider just one attestation to prune.

primitives/hp-proof-of-existence/src/lib.rs
```diff
@@ -94,3 +95,25 @@ impl sp_inherents::InherentDataProvider for InherentDataProvider {
 pub trait OnProofVerified {
     fn on_proof_verified(pubs_hash: H256);
 }
+
+/// Wrapper around u32 for MaxStorageAttestations
+#[derive(Eq, PartialEq)]
+pub struct MaxStorageAttestations(pub u32);
```
la10736, line 101: Why don't use `NonZeroU32` here?

la10736 (approved):

la10736 (approved):
LGTM

primitives/hp-proof-of-existence/src/lib.rs
DanieleDiBenedetto: Actually, I've enforced to be at least 1, otherwise it's the same effect as 0 with the current implementation

la10736 (approved):
LGTM

pallets/proof_of_existence/src/lib.rs
```diff
@@ -113,6 +120,16 @@ pub mod pallet {
...
 
             let id = Self::next_attestation();
+
+            // In case a publish_attestation() is called by root but not proofs has been inserted yet
+            // the corresponding entry is not created in the Values storage variable. To avoid this
+            // inconsistency, we manually do this.
+            if ensure_root(origin.clone()).is_ok()
+                && Values::<T>::iter_key_prefix(id).next().is_none()
+            {
+                Values::<T>::insert(id, H256::default(), ());
```
drgora, line 130: This might go unnoticed and create confusion in future?
I would suggest adding a `log::debug` line to make it a bit more readable.

pallets/proof_of_existence/src/lib.rs
```diff
@@ -121,6 +138,18 @@ pub mod pallet {
 
             Self::deposit_event(Event::NewAttestation { id, attestation });
 
+            // Prune old attestations
+            // Rationale: ids are incremental, no more than one attestation
+            // for each publish_attestation call will need to be removed from storage
+            let max_attestations = T::MaxStorageAttestations::get();
+            if max_attestations != MaxStorageAttestations::default()
+                && id + 1 >= max_attestations.into()
```
drgora, line 146: Maybe I'm missing something here, but it looks like this would likely break if we change (either way) the value of `MaxStorageAttestations` with a runtime update?
I'd say that we need to have a loop which also takes into account `oldest_attestation_id` to have a robust implementation?

primitives/hp-proof-of-existence/src/lib.rs
```diff
@@ -94,3 +95,25 @@ impl sp_inherents::InherentDataProvider for InherentDataProvider {
 pub trait OnProofVerified {
     fn on_proof_verified(pubs_hash: H256);
 }
+
+/// Wrapper around u32 for MaxStorageAttestations
```
drgora, line 99: Given that `u64` was used for attestation ids in the pallet, would it make sense to make also this a `u64` to have things more consistent?

drgora (changes requested):
Thanks for the PR; I believe there are few things worth assessing before merge, especially in the logic for the removal of old attestations, in case of changes in `MaxStorageAttestations`.

runtime/src/lib.rs
```diff
@@ -503,13 +503,24 @@ impl pallet_child_bounties::Config for Runtime {
 
 pub const MILLISECS_PER_PROOF_ROOT_PUBLISHING: u64 = MILLISECS_PER_BLOCK * 10;
 pub const MIN_PROOFS_FOR_ROOT_PUBLISHING: u32 = 16;
+pub const MAX_STORAGE_ATTESTATIONS: u32 = 100_000;
+
 // We should avoid publishing attestations for empty trees
 static_assertions::const_assert!(MIN_PROOFS_FOR_ROOT_PUBLISHING > 0);
 
+// We should keep in memory at least one attestation
+static_assertions::const_assert!(MAX_STORAGE_ATTESTATIONS > 1);
```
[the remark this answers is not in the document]
DanieleDiBenedetto: The parameter takes into account also the current attestation. So at least one must always exist in the storage

pallets/proof_of_existence/src/lib.rs
DanieleDiBenedetto: Yes. We iterated a bit on this and this is the most minimal solution that wouldn't make the weights increase too much. But I agree that it won't work if, for instance, we decrease the MaxAttestations parameter. Let me see if I can find a way to fix this in a cheaper way, but keep in mind that we will be refactoring this pallet heavily in the upcoming sprint

la10736: Just a little comment about that: we can meet this issue in just when on update runtime when the attestations in the storage are (much) bigger than  `MaxStorageAttestations`.

Now in the next upgrade we'll use a number that will be bigger than the attestation already present in the storage.

Can a more robust implementation save us later when we'll change this number drastically in a future by a runtime update?

I think the answer is NO: if we implement we cannot benchmarking the worst case but just the normal one. In this case when we issue the extrinsic and we need to remove something like 50000 attestation we have a lot of chances to kill the chain because this transaction will take all block chain to run but the estimated one is just for remove one attestation.

What I really mean here is that we cannot implement a variable behavior in our extrinsic and if we really need to do it we need to estimate the gas at least in all worst condition and in this case really we cannot do it.

What we can do is to write a migration for the pallet that will be run on runtime upgrade and without weight constrains. We should just take care of it when we'll perform an upgrade that reduce this value: my advice is to add a comment or/and (better)  add a static assert that should be greater-equal to that value to raise the red flag in the case we reduce this number.

Anyway I guess that we'll change this pallet before we'll run in this issue.

DanieleDiBenedetto: I've added more robust logic and a test case, just to close this.
Most likely we will be refactoring it in the future anyway
https://github.com/HorizenLabs/zkVerify/pull/124/commits/56cdb79994fe4bc825465bf6e4bfb848470af7bf

la10736 (approved):
My advice is that this solution try to fix an issue in a way that is worst than the issue itself. Anyway I don't think we'll never meet this issue in a future and we can leave it as it is.

pallets/proof_of_existence/src/lib.rs
drgora: Thanks

pallets/proof_of_existence/src/lib.rs
drgora: Thank you for the update. I totally agree with the comment that you added to the code, and that we should have some limits on the actual number of iterations that can be performed in reality to avoid timeout issues. Or maybe, we could consider refactoring that "cleanup" functionality into some dedicated extrinsic, to also account more predictably for the fee incurred by the extrinsic.

Anyways, given that we are already planning more changes and refactoring, I'm ok with what we have, for now.

drgora (approved):

PR #127, opened 2024-09-06
[HOLD] Relay chain impl

la10736:
Here we have 3 nodes executable

- `mainchain` crate: The solo chain `zkv-node` that should be untouched 
- `zkv-relay` crate: The relay chain `zkv-relay`
- `paratest` crate: The `paratest-node` provided as parachain example

Some istruction to run it are in `RELAY.md`.

The runtime is shared for both chain and the `relay` feature is used to enable the relay chain specific features.


runtime/src/lib.rs
```diff
@@ -781,6 +845,91 @@ construct_runtime!(
...
+        ParasDisputes: parachains::disputes = 112,
+        ParasSlashing: parachains::slashing = 113,
+        ParachainsAssignmentProvider: parachains::parachains_assigner_parachains = 115,
+
+        // Parachain chain (removable) pallets. Start indices at 130.
+        ParasSudoWrapper: parachains::paras_sudo_wrapper = 130,
+
+        // XCM Pallet: start indices at 140.
+        XcmPallet: pallet_xcm = 140,
+        MessageQueue: pallet_message_queue = 114,
```
drgora, line 929: Is there any reason for this pallet to have value `114`? Maybe we can make it `141`, and at the same time assign `ParachainsAssignmentProvider = 114`?

drgora (approved):
Looks good to me!

runtime/src/lib.rs
la10736: Yeah... I just forgot it!

DanieleDiBenedetto (approved):

DanieleDiBenedetto:
We should update the docker.sh passing as additional argument to the build `-p mainchain` @drgora

DanieleDiBenedetto:
Closed in favor of #163

PR #128, opened 2024-09-06
feat: Add pallet bags list

rdoria1:
The `runtime/src/bag_thresholds.rs` should be autogenerated (similar to the weight file) for our network.

generate bags script added with following command line

`cargo run -p node-runtime-generate-bags -- --node-url wss://testnet-rpc.zkverify.io runtime/bag_thresholds.rs`


la10736:
I think you should also add here the `generate_bags`: in this case isn't relevant if this code run on reference machine or not because the result depends only by how our runtime is configured and the values used to generate it (`n_bags`, `total_issuance` and `minimum_balance`).

We can also chose to take `minimum_balance == existential_deposit` and remove it from the command line values and make the process less error prone by read it from runtime. Maybe for us could be also useful set a default for `n_bags` and provide an url to read the total issuance (where the default value point to testnet).

Just to summarize:
- copy the code in https://github.com/paritytech/polkadot-sdk/tree/master/substrate/utils/frame/generate-bags in our `utils` folder. 
- Adapt the `node-runtime` code to use our runtime 
- Get the default value for `minimum_balance` from the e`existential_deposit` 
- (Optional) read the total issuance from the given chain

utils/generate-bags/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+homepage = "https://substrate.io"
+repository.workspace = true
+description = "Bag threshold generation script for pallet-bag-list"
+
+[dependencies]
+frame-support = { default-features = false, version = "32.0.0" }
+frame-election-provider-support = { default-features = false, version = "32.0.0" }
+frame-system = { default-features = false, version = "32.0.0" }
+pallet-staking = { default-features = false, version = "32.0.0" }
+sp-staking = { default-features = false, version = "30.0.0" }
```
drgora, line 16: Can we use the `workspace` versions here, as we do in the rest of the project?

utils/generate-bags/node-runtime/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition.workspace = true
+license = "Apache-2.0"
+homepage = "https://substrate.io"
+repository.workspace = true
+description = "Bag threshold generation script for pallet-bag-list and zkv-runtime."
+publish = false
+
+[dependencies]
+zkv-runtime = { version = "0.5.1", path = "../../../runtime" }
+generate-bags = { version = "32.0.0", path = ".." }
```
drgora, line 14: I'd suggest to skip the version tag (which is implicit) when you pass a local path

utils/generate-bags/src/lib.rs
```diff
@@ -0,0 +1,277 @@
+// Copyright 2024, Horizen Labs, Inc.
+
```
DanieleDiBenedetto, line 2: Shouldn't we add copyright header for Parity here as well  ?

utils/generate-bags/node-runtime/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+        .await?
+        .fetch(&zk_verify::storage().balances().total_issuance())
+        .await?
+        .unwrap_or_default();
+
+    generate_thresholds::<zkv_runtime::Runtime>(
+        n_bags,
+        &output,
+        total_issuance,
+        zkv_runtime::EXISTENTIAL_DEPOSIT,
```
drgora, line 63: I believe it would be better to fetch the `existential_deposit` as a constant from the given RPC, similarly to what you do with the `total_issuance`, instead of taking it from the local code.

utils/generate-bags/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/// Compute the list of bag thresholds.
+///
+/// Returns a list of exactly `n_bags` elements, except in the case of overflow.
+/// The first element is always `existential_weight`.
+/// The last element is always `VoteWeight::MAX`.
+///
+/// All other elements are computed from the previous according to the formula
+/// `threshold[k + 1] = (threshold[k] * ratio).max(threshold[k] + 1);`
```
DanieleDiBenedetto, line 128: I assume yes, but please double check with John if this way of computing the bags thresholds is fine.

utils/generate-bags/node-runtime/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+use generate_bags::generate_thresholds;
+use std::path::PathBuf;
+use subxt::{OnlineClient, SubstrateConfig};
+
+#[derive(Debug, Parser)]
+// #[clap(author, version, about)]
+struct Opt {
+    /// How many bags to generate.
+    #[arg(long, default_value_t = 200)]
+    n_bags: usize,
```
DanieleDiBenedetto, line 28: Let's probably agree with John how to set this value

drgora (changes requested):
Thank you for adding the new pallet. The main thing that I believe we should do before merging, is to actually use the new pallet to keep track of the voters.
This should be done in the configuration of the Staking pallet. I also left few other comments that might be worth addressing.

DanieleDiBenedetto:
> I think you should also add here the `generate_bags`: in this case isn't relevant if this code run on reference machine or not because the result depends only by how our runtime is configured and the values used to generate it (`n_bags`, `total_issuance` and `minimum_balance`).
> 
> We can also chose to take `minimum_balance == existential_deposit` and remove it from the command line values and make the process less error prone by read it from runtime. Maybe for us could be also useful set a default for `n_bags` and provide an url to read the total issuance (where the default value point to testnet).
> 
> Just to summarize:
> 
> * copy the code in https://github.com/paritytech/polkadot-sdk/tree/master/substrate/utils/frame/generate-bags in our `utils` folder.
> * Adapt the `node-runtime` code to use our runtime
> * Get the default value for `minimum_balance` from the e`existential_deposit`
> * (Optional) read the total issuance from the given chain

Not a big fan of tying this tool to an actual running/configured chain. I see Polkadot is exposing these values as input parameters for the CLI perhaps it's better to do in this way (so one can also do some local experiments if desired), or maybe encoding a logic s.t., if they are not specified, the code attempts to read it from the chain ?

runtime/src/lib.rs
```diff
@@ -272,6 +273,21 @@ impl pallet_babe::Config for Runtime {
...
+
+type VoterBagsListInstance = pallet_bags_list::Instance1;
+type VoteWeight = u64;
+
+impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type ScoreProvider = Staking;
+    type WeightInfo = weights::pallet_bags_list::ZKVWeight<Runtime>;
+    type BagThresholds = BagThresholds;
+    type Score = VoteWeight;
```
DanieleDiBenedetto, line 288: Shouldn't this be taken from the election pallet @drgora ?

DanieleDiBenedetto:
Also, let's not forget to run the benches for the staking pallet, now that we have integrated the bags list pallet
https://github.com/HorizenLabs/zkVerify/blob/main/scripts/run_all_benchmarks.sh#L109

utils/generate-bags/node-runtime/src/main.rs
rdoria1: Unfortunately I couldn't  retrieve it from API nor RPC

I found two "potential" ways to get the value, but none of them work (probably something is not implemented)

-> https://polkadot.js.org/docs/substrate/runtime#queryexistentialdepositcurrencyid-raw-u128 but tokens API is not implemented in the API

-> Following from integ tests in subxt https://github.com/paritytech/subxt/blob/9739a1682c5e81c7bfe539a08c3c1b0fb3e4798f/testing/integration-tests/src/full_client/metadata/validation.rs#L122
Same, doesn't work 

I'll keep it as it is for now.

utils/generate-bags/src/lib.rs
rdoria1: asked John, waiting on reply

utils/generate-bags/node-runtime/src/main.rs
rdoria1: asked John, waiting on reply

utils/generate-bags/node-runtime/src/main.rs
drgora: Ok but this is weird, I would expect the latter to work on our testnet, the constants are accessible from polkadot-js, and there's also an example code that looks quite similar to yours: https://github.com/paritytech/subxt/blob/9739a1682c5e81c7bfe539a08c3c1b0fb3e4798f/subxt/examples/constants_static.rs

We should investigate this at some point.

runtime/src/lib.rs
drgora: Oh yes, good catch. `VoteWeight` would be better be taken from `sp-npos-elections`, which we already import indirectly.

utils/generate-bags/src/lib.rs
rdoria1: this is fine, confirmed

utils/generate-bags/node-runtime/src/main.rs
rdoria1: 200 is fine for now, asked John

utils/generate-bags/node-runtime/src/main.rs
rdoria1: @drgora 
I'll put a ticket on the backlog

utils/generate-bags/src/lib.rs
rdoria1: Added both as suggested by @la10736 (synced offline)

DanieleDiBenedetto (approved):

drgora (approved):
LGTM, thanks!

PR #131, opened 2024-09-11
We should avoid to repeate the `${VAR:-}` patter in the script

la10736:


DanieleDiBenedetto (approved):

JackPiri (approved):
The PR is fine to me.
Thanks

PR #132, opened 2024-09-17
Use polkadot crypto from zombie

la10736:
We should not import @polkadot/util-crypto again because is already exported in zombie

https://github.com/paritytech/zombienet/blob/e9f0993f15ed4a8d9ad1589056ea188ed1b0d039/javascript/packages/orchestrator/src/test-runner/assertions.ts#L303C13-L303C23


DanieleDiBenedetto (approved):

drgora (approved):

la10736:
Please ignore this PR!!!! I've definitely broke it

PR #133, opened 2024-09-18
Integration tests on parachains and some e2e test on relay chain

la10736:
We add some integration tests about parachain's pallets weights and some e2e tests on relay.

A notable e2e test is the one that use horizen chain spec to register horizen parachain and forge some blocks.


DanieleDiBenedetto (approved):

drgora (approved):
LGTM, thanks!

PR #134, opened 2024-09-18
Introduce custom inflation model

drgora:
The title says it all; this PR introduces a new custom inflation model, based on a variable rate which depends on the staking rate at a given point.

For this, an `exp()` function is introduced. Tests are provided


runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
+use super::*;
+pub use sp_runtime::{Percent, Perquintill};
+
+fn abs(v: f64) -> f64 {
```
DanieleDiBenedetto, line 4: Do you envision having negative exponents ?

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+        if total_issuance == 0 {
+            return (0, 0);
+        }
+
+        const MILLISECS_PER_YEAR: u64 = 1000 * 60 * 60 * 24 * (36525 / 100);
+        let time_portion = Perquintill::from_rational(era_duration_millis, MILLISECS_PER_YEAR);
+
+        let staking_current: Perbill = Perbill::from_rational(total_staked, total_issuance);
+        let s = StakingTarget::get()
+            .saturating_reciprocal_mul(staking_current * (1f64 / ExpPrecision::get()) as u128);
```
la10736, line 79: Add a comment here to map s into the formula: `s = s_c/s_t`

la10736: Also in the next two lines

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+        let exp_arg = ((1f64 / ExpPrecision::get()) as i128 - s as i128) * K::get();
```
DanieleDiBenedetto, line 80: Unless we pick a negative K (why ?) the exponent can never be negative

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+            C::get() * (time_portion * (InflationBase::get() + inflation_var) * total_issuance);
+
+        (
+            ValidatorsSplit::get() * inflation_tot,
+            OthersSplit::get() * inflation_tot,
+        )
+    }
+}
+
+/// This module tests the correct computation of rewards for validators.
```
la10736, line 93: Ok, move this comment at the top of the file and use `//!` to indicate a global comment: here you're attaching the comment to `EraPayout` item

runtime/src/payout.rs
DanieleDiBenedetto, line 79: Maybe I'm wrong, but reading the doc it seems that this function rounds the result to the nearest whole (I assume that by "whole" it means integer) number. Is this what we want ?

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+#[cfg(test)]
+use pallet_staking::EraPayout;
+
```
la10736, line 96: Move all test stuff in a module gated by `test` config: call it `test` should be fine in this contests because all tests speaking about to check something`

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        const MILLISECS_PER_YEAR: u64 = 1000 * 60 * 60 * 24 * (36525 / 100);
+        let time_portion = Perquintill::from_rational(era_duration_millis, MILLISECS_PER_YEAR);
+
+        let staking_current: Perbill = Perbill::from_rational(total_staked, total_issuance);
+        let s = StakingTarget::get()
+            .saturating_reciprocal_mul(staking_current * (1f64 / ExpPrecision::get()) as u128);
+        let exp_arg = ((1f64 / ExpPrecision::get()) as i128 - s as i128) * K::get();
+        let inflation_var = to_inflation(exp_arg as f64 / (1f64 / ExpPrecision::get()));
```
DanieleDiBenedetto, line 81: Maybe redundant conversion here consider what you do in the `to_inflation` function ?

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+
+#[test]
+fn check_params_sanity() {
+    // staking_target should be at least 0.5 (and so 0 <= s_c / s_t <= 2)
+    assert!(StakingTarget::get() * 10u16 >= 5u16);
+    // base inflation is 5%
+    assert!(InflationBase::get() * 100u64 == 5u64);
+    // The sum of the splits is exactly 100%
+    assert!(ValidatorsSplit::get() * 100u16 + OthersSplit::get() * 100u16 == 100u16);
+}
```
la10736, line 108: Do we have some reference about why these numbers are _correct_ and what are the valid range in any case?

The second assert seams to say any other number is just wrong. Maybe can be enough to be explicit in the assert message to alert the developer where to look to check if the new value is correct.

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+        let time_portion = Perquintill::from_rational(era_duration_millis, MILLISECS_PER_YEAR);
+
+        let staking_current: Perbill = Perbill::from_rational(total_staked, total_issuance);
+        let s = StakingTarget::get()
+            .saturating_reciprocal_mul(staking_current * (1f64 / ExpPrecision::get()) as u128);
+        let exp_arg = ((1f64 / ExpPrecision::get()) as i128 - s as i128) * K::get();
+        let inflation_var = to_inflation(exp_arg as f64 / (1f64 / ExpPrecision::get()));
+
+        let inflation_tot: Balance =
+            C::get() * (time_portion * (InflationBase::get() + inflation_var) * total_issuance);
```
DanieleDiBenedetto, line 84: Shouldn't `C` multiply only the `inflation_var` ?

la10736 (changes requested):
Nice job.

Just dettails

DanieleDiBenedetto (approved):
LGTM overall, left a couple of comments :)

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+/// Implements a custom inflation model based on the following formula:
+/// I(s_c) = I_b + I_v(s_c)
+///
+/// with:
+/// I_v(x) = C * exp(K * (1 - (s_c / s_t)))
+///
+/// where:
+/// I_b: base inflation
+/// I_v: variable inflation
+/// s_t: target inflation
```
la10736, line 41: that's target _staking_ and not inflaction

la10736: Maybe is good to define the target inflation `I_t` and the base inflation `I_b` and then compute `C = I_t - I_b`: I think it'll be more clear.

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+    fn era_payout(
+        total_staked: Balance,
+        total_issuance: Balance,
+        era_duration_millis: u64,
+    ) -> (Balance, Balance) {
+        if total_issuance == 0 {
+            return (0, 0);
+        }
+
+        const MILLISECS_PER_YEAR: u64 = 1000 * 60 * 60 * 24 * (36525 / 100);
```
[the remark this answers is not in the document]
la10736: NICE CATCH!!!!

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+    if v > 0f64 {
+        v
+    } else {
+        -v
+    }
+}
+
+fn exp(p: f64) -> f64 {
+    let mut res: f64 = 0f64;
+    // Sums: 1 + x + x^2/2! + x^3/3! + x^4/4! + x^5/5! +
```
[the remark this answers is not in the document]
la10736: `p` is always lower than 1. For the negative values we can just set a cut off when the current staking is less `1/21` of the target one. In this case the _correction_ of the target inflation is since `C*0.000000002` percent => we can fall to use the base inflation.

runtime/src/payout.rs
drgora: Good point, done

runtime/src/payout.rs
drgora: Yeah, refactored a bit also considering the other comments

runtime/src/lib.rs
```diff
@@ -597,6 +583,8 @@ impl pallet_staking::BenchmarkingConfig for ElectionProviderBenchmarkConfig {
     type MaxValidators = ConstU32<MAX_TARGETS>;
 }
 
+mod payout;
+
```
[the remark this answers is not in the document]
drgora: Moved it next to the others

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
+use super::*;
```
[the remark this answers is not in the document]
drgora: Right, done!

runtime/src/payout.rs
drgora: Added comments

runtime/src/payout.rs
drgora: Done

runtime/src/payout.rs
drgora: Ok, moved at the beginning of the test module

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+    // Check the reward for a normal era, s_c == s_t
+    let tot_issuance: Balance = TOT_ISSUANCE;
+    let tot_staked: Balance = Perbill::from_percent(50) * tot_issuance;
+    assert_eq!(
+        <Runtime as pallet_staking::Config>::EraPayout::era_payout(
+            tot_staked,
+            tot_issuance,
+            ERA_DURATION_MILLIS
+        ),
+        (60000000 * ACME, 0)
```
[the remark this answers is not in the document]
drgora: Precisely, that was also my reasoning.

runtime/src/payout.rs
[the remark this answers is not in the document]
drgora: Definitely forgot this, thanks!

runtime/src/payout.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// The target precions for exp(); impacts on the final precision for inflation computation.
+    pub const ExpPrecision: f64 = 10e-15f64;
+    /// Base inflation (I_b).
+    pub InflationBase: Perquintill = Perquintill::from_percent(5);
+    /// The optimal staking rate (s_t).
+    pub StakingTarget: Percent = Percent::from_percent(50);
+    /// Sensitivity coefficient (K).
+    pub const K: i128 = 1i128;
+    /// Multiplier (C).
+    pub const C: u128 = 1u128;
```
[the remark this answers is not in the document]
drgora: It's not entirely clear what values these will have in the end, made them `f64` (could be > 1) and `Perquintill` respectively, for convenience.

runtime/src/payout.rs
drgora: For now I would stick with the current representation, where the _total_ inflation `I_t` is variable, based on the current staking rate `s_c`.
This is closer to the "theoretical" description.

runtime/src/payout.rs
```diff
@@ -24,7 +46,7 @@ fn exp(p: f64) -> f64 {
 
 fn to_inflation(p: f64) -> Perquintill {
     Perquintill::from_rational(
-        (exp(p) * (0.01f64 / ExpPrecision::get())) as u64,
+        (p * (0.01f64 / ExpPrecision::get())) as u64,
         (1f64 / ExpPrecision::get()) as u64,
```
DanieleDiBenedetto, line 50: Apply `SCALE` also here ?

DanieleDiBenedetto (approved):

runtime/src/payout.rs
la10736: :+1:

runtime/src/payout.rs
la10736: :+1:

la10736 (approved):
LGTM THX

PR #138, opened 2024-09-24
Staking benchmark

drgora:
This PR updates the runtime configuration to fix the benchmarks for the staking pallet, while still using safe values for on chain elections.

The PR also adds ZKVWeight for Staking, as measured on our reference machine.


runtime/src/lib.rs
```diff
@@ -569,13 +569,18 @@ parameter_types! {
...
+pub const MAX_VOTERS: u32 = 5_000;
 
 parameter_types! {
-    pub ElectionBoundsOnChain: ElectionBounds = ElectionBoundsBuilder::default().voters_count((MAX_TARGETS + MAX_VOTERS).into()).targets_count(MAX_TARGETS.into()).build();
+    // Maximum number of election voters and targets that can be handled by OnChainSeqPhragmen
+    // in a single block time.
+    pub ElectionBoundsOnChain: ElectionBounds = ElectionBoundsBuilder::default().voters_count((MAX_VOTERS).into()).targets_count(MAX_TARGETS.into()).build();
+    // Maximum number of election winners, and thus of authorities that can be active in a given
+    // era.
     pub const MaxActiveValidators: u32 = MAX_TARGETS;
```
DanieleDiBenedetto, line 584: Is it always true that the maximum number of active validators equals the maximum number of eligible targets ? Are there bounds enforced somewhere ?

DanieleDiBenedetto (approved):

runtime/src/lib.rs
la10736: Yes, that's enforced by `DataProvider` configuration (staking pallet) that provide max `MAX_TARGET` targets on `desired_targets()` call.

la10736 (approved):
LGTM

PR #141, opened 2024-09-30
Migrate storage to bags list

drgora:
This PR adds some code to handle the migration of voters from the previously used `pallet_staking::UseNominatorsAndValidatorsMap` to bags list  (see #128).

Additionally, this PR moves `VoterList` to the end of the list for construct_runtime, to preserve the previous ids for the other pallets (and avoid storage issues with `Balances`). Notice: this is a temporary solution, which will be succeeded by hardcoded indices in the relay chain PR.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -808,7 +808,7 @@ pub type SignedExtra = (
 ///
 /// This can be a tuple of types, each implementing `OnRuntimeUpgrade`.
 #[allow(unused_parens)]
-type Migrations = ();
+type Migrations = migrations::Unreleased;
 
 /// Unchecked extrinsic type as expected by this runtime.
 pub type UncheckedExtrinsic =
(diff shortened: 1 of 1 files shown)


runtime/src/lib.rs
```diff
@@ -871,6 +871,78 @@ pub const BABE_GENESIS_EPOCH_CONFIG: sp_consensus_babe::BabeEpochConfiguration =
...
+
+                    log::info!(
+                        "👜 completed staking migration to bags list with {} voters migrated",
+                        migrated,
+                    );
+
+                    BlockWeights::get().max_block
+                } else {
+                    <Runtime as frame_system::Config>::DbWeight::get().reads(1)
+                }
```
la10736, line 923: I prefer the _guard_ approach instead... But feel free to leve it as it is.

la10736 (approved):
LGTM

DanieleDiBenedetto (approved):

runtime/src/lib.rs
drgora: Thanks for the tip, I'll look into that for future reference. Anyways, this is what Polkadot did for the equivalent migration and it looks reasonable in this case, I'd leave it as it is this time.

PR #142, opened 2024-10-01
Add proxy pallet

DanieleDiBenedetto:


runtime/src/lib.rs
```diff
@@ -778,6 +805,7 @@ construct_runtime!(
         Bounties: pallet_bounties,
         ChildBounties: pallet_child_bounties,
         Utility: pallet_utility,
+        Proxy: pallet_proxy,
```
la10736, line 808: Till we don't use the index is better to add new pallets at the end of the list.

runtime/src/proxy.rs
```diff
@@ -0,0 +1,120 @@
...
+                                                            // RuntimeCall::Registrar(paras_registrar::Call::register {..}) |
+                                                            // RuntimeCall::Registrar(paras_registrar::Call::deregister {..}) |
+                                                            // // Specifically omitting Registrar `swap`
+                                                            // RuntimeCall::Registrar(paras_registrar::Call::reserve {..}) |
+                                                            // RuntimeCall::Crowdloan(..) |
+                                                            // RuntimeCall::Slots(..) |
+                                                            // RuntimeCall::Auctions(..) | // Specifically omitting the entire XCM Pallet
+                                                            // RuntimeCall::NominationPools(..) |
+                                                            // RuntimeCall::FastUnstake(..) |
+                                                            // RuntimeCall::Claims(..)
```
la10736, line 73: Please remove commented lines: if somebody'll forget to handle the Proxy conf when add one of this pallets these commented lines will not help him

runtime/src/proxy.rs
```diff
@@ -0,0 +1,120 @@
...
+                    | RuntimeCall::Whitelist(..)
+            ),
+            ProxyType::Staking => {
+                matches!(
+                    c,
+                    RuntimeCall::Staking(..)
+                        | RuntimeCall::Session(..)
+                        | RuntimeCall::Utility(..)
+                        | RuntimeCall::VoterList(..) // RuntimeCall::FastUnstake(..) |
+                                                     // RuntimeCall::NominationPools(..)
```
la10736, line 100: Remove commented code

runtime/src/proxy.rs
```diff
@@ -0,0 +1,120 @@
...
+                )
+            }
+        }
+    }
+    fn is_superset(&self, o: &Self) -> bool {
+        match (self, o) {
+            (x, y) if x == y => true,
+            (ProxyType::Any, _) => true,
+            (_, ProxyType::Any) => false,
+            (ProxyType::NonTransfer, _) => true,
```
la10736, line 116: That seams the most tricky rule because it claims that all no `Any` poxies should be a sub set of `NonTransfer`.

It should be nice if we can find a way to test it but I cannot see how to find a simple way to create all possible transactions in the other proxies.

la10736 (approved):
LGTM

All my commnets are optional. Feel free to ignore them.

runtime/src/proxy.rs
drgora: We might work around the problem expanding line 116 into explicit cases (e.g. `(ProxyType::NonTransfer, ProxyType::Governance) => true`, ...).
That would make this safer against future additions of ProxyTypes at the cost of 2 additional lines, but honestly I don't like this option much.

Even simpler, we might just add a comment between lines `21-22` saying not to add any values in that position :)

drgora (approved):
LGTM, thanks!

PR #145, opened 2024-10-04
Md/groth16_native

la10736:
Port to groth16 native implementation .

Benchmarking evidentiate a >x5 boost

Introduced a primitive crate `hp-groth16` that's the only one that depends by arkworks crates. This crate expose its own types for both native and runtime parts.


DanieleDiBenedetto (approved):
Amazing work !

PR #146, opened 2024-10-04
Rename e2e tests to zombienet tests

yurikoinaba:


la10736:
Hi @yurikoinaba , can you provide a little context about this PR?
I'm not against it, but I would like to understand why we should rename it. I don't like to use _technology name_ instead of _aim name_. Maybe `e2e_tests` is not the best choice but I prefer to know why `zombienet_tests` is better.

la10736:
Hi @yurikoinaba ,
To fix the CI you have 2 options:
1. create the folder `zombienet-tests/bin` and put an empty `.placeholder` file. To commit it you should temporally remove `zombienet-tests/bin` from `.gitignore` and after add it again
2. (maybe better) add a `mkdir -p bin` command after line 46 in `zombienet-tests/test-runner.sh` script

yurikoinaba:
@la10736 Thank you, I'll update the test-runner.sh script

I'd like to rename these tests to avoid confusion w[ the e2e in the zkverify-qa repo](https://github.com/HorizenLabs/zkVerify-qa/tree/main/src/e2e-tests). There will be an update to run the zkverify-qa e2e tests on tag creation. 

Instead of `zombienet-test`, it can also be renamed to `systems-test` or `network-test`

la10736 (approved):
LGTM

ysibirski (approved):
LGTM

PR #147, opened 2024-10-10
Use groth16 WASM implementation instead of the native one

la10736:
Just separate the feature std from _implementaion_ that provide the curves implementation in the `hp-groth16` crate.

Now the verifier use the `hp-groth16` instead the native one so the node proviede the native implementation for the future, but the verifier not. 

Also I make the verifier functions interface all the same.

Missing: rerun the benchmarks


primitives/hp-groth16/Cargo.toml
```diff
@@ -39,7 +39,8 @@ rstest_reuse = { workspace = true }
 hex-literal = { workspace = true }
 
 [features]
-default = ["std"]
+default = ["std", "implementation"]
+implementation = []
```
DanieleDiBenedetto, line 43: Maybe we can find a better name ? Something like "use-wasm" ?

DanieleDiBenedetto (approved):

DanieleDiBenedetto:
Would you be able to test locally ? Like try to sync to the testnet a node with new node and runtime version, submit a Groth16 proof ?

la10736:
I did the follow tests:
1. Started a new local chain using the docker compose file provided in the source after pulled the `horizenlabs/zkverify:latest` docker image from docker hub, performed a runtime upgrade and checked that the update is ok and that the chain continues to forge new blocks
2. A new synch of the testenet chain (make sure that I'm not use an old cache)

Bonus: In order to be sure that I'm able to spot issue I generated a runtime that need the native groth16 implementation and repeated 1; in this case the runtime update executed successfully  but the chain stop to generate new blocks.

la10736:
About submitting a new groth16 proof we have the e2e tests that protect us by proof submission regression.

primitives/hp-groth16/Cargo.toml
la10736: From this crate point of view it's not `use-wasm` but is provide or not a groth16 implementation. I don't like `implementation` too but I didn't find any better options. Some options

1. `arkworks-impl`
2. `verify-impl`
3. `verify`

lgiussan (approved):
LGTM, thank you

PR #148, opened 2024-10-14
Integrate xcm configuration and benchmarks

drgora:


runtime/src/xcm_config.rs
```diff
@@ -510,6 +244,8 @@ impl xcm_executor::Config for XcmConfig {
     type HrmpChannelClosingHandler = ();
 }
 
+const FELLOWSHIP_ADMIN_INDEX: u32 = 1; // to be moved to some constants mod
```
DanieleDiBenedetto, line 247: What is this ? Is this mandatory ?

DanieleDiBenedetto (approved):
Thanks for the amazing work !

runtime/src/xcm_config.rs
drgora: No it isn't, as we are not using any of these as of today. I also liked the idea of removing this, but then I thought to keep it consistent with the rest of the runtime, that is with the _governance_ part in this case.

relay-node/tests/purge_chain_works.rs
```diff
@@ -36,7 +36,7 @@ const PARITY: &str = "paritydb/full";
 
 #[tokio::test]
 async fn purge_chain_rocksdb_works() {
-    run_with_timeout(Duration::from_secs(10 * 60), async move {
+    run_with_timeout(Duration::from_secs(20 * 60), async move {
```
la10736, line 39: Maybe 5 is enough

relay-node/tests/purge_chain_works.rs
```diff
@@ -80,7 +80,7 @@ async fn purge_chain_rocksdb_works() {
 
 #[tokio::test]
 async fn purge_chain_paritydb_works() {
-    run_with_timeout(Duration::from_secs(10 * 60), async move {
+    run_with_timeout(Duration::from_secs(20 * 60), async move {
```
la10736, line 83: See above comment

runtime/src/lib.rs
```diff
@@ -369,19 +370,21 @@ impl pallet_timestamp::Config for Runtime {
...
-pub const EXISTENTIAL_DEPOSIT: u128 = MILLICENTS;
+parameter_types! {
+    /// Existential deposit.
+    pub const ExistentialDeposit: Balance = EXISTENTIAL_DEPOSIT;
+}
 
 impl pallet_balances::Config for Runtime {
     type MaxLocks = ConstU32<50>;
-    type MaxReserves = ();
+    type MaxReserves = ConstU32<50>;
```
la10736, line 380: Just curiosity ... why?

la10736 (approved):
LGTM.... Just reduce the test timeout to 5 minutes

relay-node/tests/purge_chain_works.rs
drgora: Right, let's make it `5`

relay-node/tests/purge_chain_works.rs
drgora: Done

runtime/src/lib.rs
drgora: I was using this to test some `xcm` scenarios. Anyways, I will add some e2e tests to cover the basic xcm interactions that we want to support, and then remove this if unneeded.

PR #150, opened 2024-10-15
Disable runtime migration for bags list

drgora:
This PR disables the migration code for the previous runtime upgrade (`5_002` -> `6_000`).
The disabled code is moved to a specific file, and kept for clarity and future reference.


DanieleDiBenedetto (approved):
Just out of curiosity what would happen if we left the code active ?

la10736 (approved):
LGTM

PR #152, opened 2024-10-18
Disable verifier feature

la10736:
Rational: we need a way to handle issues related to verifiers that have bugs or wrongly weights that can cause chains attacks.

- Implement a storage variable for, eventually,  disabling a verifier.
- Just root can disable/enable a verifier.
- Verify on a disable verifier should cost just the check and not all the weight
-  Introduced a common pallet to handle the weights that are the same for all verifiers (enable/disable and pay for a verification on a disable verifier)

Future works:
- Introduce a verifier manager for all verifiers and eventually one for each verifier instead of use the hardcoded root user only (if we want to remove sudo).
- Write a benchmark for the common weights instead of use the hardcoded implementation.

### DIFF
--- a/pallets/verifiers/src/tests.rs
+++ b/pallets/verifiers/src/tests.rs
@@ -22,6 +22,7 @@ use sp_core::H256;
 
 use super::*;
 use crate::mock::*;
+use frame_support::{assert_err, assert_err_ignore_postinfo};
 use rstest::{fixture, rstest};
 
 type Vk = <FakeVerifier as Verifier>::Vk;
@@ -32,6 +33,8 @@ type VkOrHash = super::VkOrHash<Vk>;
 pub fn test_ext() -> sp_io::TestExternalities {
     crate::mock::test_ext()
 }
+
+type DisableStorage = Disabled<Test, FakeVerifier>;
 
 mod register_should {
     use hex_literal::hex;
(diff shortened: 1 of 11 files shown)


pallets/verifiers/src/tests.rs
lgiussan, line 1: I would add a unit test which ensures that it's possible to successfully verify a valid proof when the `Disabled` storage value is unset (i.e. before any call to the `disable` extrinsic has ever been performed)

lgiussan (commented):
Great, LGTM! Just a suggestion about an additional test

pallets/verifiers/src/tests.rs
lgiussan: Nevermind, [this test](https://github.com/HorizenLabs/zkVerify/blob/07281b28240f040fb184b431464d71e64fe061b4/pallets/verifiers/src/tests.rs#L117) already ensures it. Perhaps you could explicitly add

```rust
assert_eq!(FakeVerifierPallet::disabled(), None);
```
inside the test.

lgiussan (approved):

pallets/verifiers/src/tests.rs
la10736: I'm not sure that is better.

1. There are lot of tests that check that in the _normal_ contidions the verification works.
2. There is a test that ensure if you disable the verifier the verification raise error
3. There is a test that ensure if you enable a verifier the it was previously disabled the verification works

I'm trying to isolate tests as much as possible to check less precondition as possible and minimize the number of asserts in the test. I understood that you would be sure that this behaviour is checked, but I cannot see any, let me say, _trivial_ code modification that can lead to an involutary regression that will not spotted out by these tests.

lgiussan (commented):
On a second thought, maybe it's worthwhile to disable also vk registration. What do you think? It shouldn't require much additional code/tests

rdoria1 (approved):
LGTM

la10736:
> On a second thought, maybe it's worthwhile to disable also vk registration. What do you think? It shouldn't require much additional code/tests

I had thought about this before I my first idea was to not implement it . But I changed my mind and implement it.

PR #153, opened 2024-10-18
Coverage with latest rust version

la10736:
Exclude mainchain from coverage to workaround the compiler issue and recover the use of the last toolchain.

Till https://github.com/rust-lang/rust/issues/125353 is not fixed we should exclude `mainchain` from our coverage that cause a compilation panic.


lgiussan (approved):
LGTM

DanieleDiBenedetto (approved):

PR #160, opened 2024-10-23
Skip compilation of script for generating bags-thresholds

DanieleDiBenedetto:
Compilation of the binary takes quite a bit of time actually. Since it's something used once in a while (perhaps not even worth to be committed) I just commented the corresponding line in the workspace members as a not-too-cumbersome solution.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,9 +21,12 @@ members = [
     "verifiers/zksync",
     "verifiers/groth16",
     "verifiers/risc0",
-    "verifiers/ultraplonk", 
+    "verifiers/ultraplonk",
     "verifiers/proofofsql",
     "utils/native-cache",
+]
+
+exclude = [
     "utils/generate-bags/node-runtime",
 ]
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -24,7 +24,8 @@ members = [
     "verifiers/ultraplonk", 
     "verifiers/proofofsql",
     "utils/native-cache",
-    "utils/generate-bags/node-runtime",
+    # Uncomment when in need to re-generate bags thresholds
+    # "utils/generate-bags/node-runtime",
```
la10736, line 28: I don't like this comment and also is not necessary: even without uncomment it you can go in `utils/generate-bags/node-runtime` and run `cargo build` (I guess you need just replace the workspace references in `Cargo.toml`).

Anyway is just a mater of taste.

la10736 (approved):
LGTM

drgora (approved):
LGTM

Cargo.toml
DanieleDiBenedetto: You'd need to update the Cargo.toml removing all references to workspace. Doable but then you need to duplicate dependencies and you need to keep them up-to-date. Probably better to just uncomment the line in the root Cargo.toml if needed as to avoid possible headaches/inconsistencies, if you agree

la10736 (approved):

PR #161, opened 2024-10-24
Update polkadot 1.15.0

rdoria1:
**Pallet treasury**
remove _ApproveOrigin_ -> Found this comment on this type before it was effectively removed  https://github.com/paritytech/polkadot-sdk/pull/4831/files
// The creation of proposals via the treasury pallet is deprecated and should not be utilized.
 	// Instead, public or fellowship referenda should be used to propose and command the treasury
 	// spend or spend_local dispatchables. The parameters below have been configured accordingly to
 	// discourage its use.
Remove _OnSlash_-> This was added to the bounty pallets instead (PR comment in https://github.com/paritytech/polkadot-sdk/pull/4831)
Remove _ProposalBond_ _ProposalBondMinimum_ and _ProposalBondMaximum_ -> https://github.com/paritytech/polkadot-sdk/pull/3820/commits/3d26a9235d84fb3b353ac537e4c1b49d482581a1 I feel this is because they all related to proposal and based on the first point. proposals are not done via the treasury pallet anymore which is why these types are useless now

**Pallet bounties**
Added _OnSlash_ as specified before, coming from treasury

**Pallet staking**
Replace _OffendingValidatorsThreshold_ with _DisablingStrategy_ https://github.com/paritytech/polkadot-sdk/pull/2226

**node**
remove deprecated _RuntimeGenesisConfig_ : https://github.com/paritytech/polkadot-sdk/pull/4410

**TODO** Re run benchmark for **pallet_balances**

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -118,7 +118,7 @@ pub fn authority_ids_from_ss58(
 pub fn development_config() -> Result<ChainSpec, String> {
     Ok(ChainSpec::builder(
         WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?,
-        Default::default(),
+        None,
     )
     .with_name("Development")
     .with_id("dev")
@@ -153,7 +153,7 @@ pub fn development_config() -> Result<ChainSpec, String> {
 pub fn local_config() -> Result<ChainSpec, String> {
     Ok(ChainSpec::builder(
         WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?,
-        Default::default(),
+        None,
     )
     .with_name("ZKV Local")
     .with_id("zkv_local")
@@ -194,7 +194,7 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
 pub fn testnet_config_build() -> Result<ChainSpec, String> {
     Ok(ChainSpec::builder(
         WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?,
-        Default::default(),
+        None,
     )
     .with_name("ZKV Testnet")
     .with_id("zkv_testnet")
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -21,7 +21,7 @@ use sp_consensus_grandpa::AuthorityId as GrandpaId;
 use sp_core::{sr25519, Pair, Public};
 use sp_runtime::traits::{IdentifyAccount, Verify};
 use zkv_runtime::currency::{Balance, ACME};
-use zkv_runtime::{currency, AccountId, RuntimeGenesisConfig, SessionKeys, Signature, WASM_BINARY};
+use zkv_runtime::{currency, AccountId, SessionKeys, Signature, WASM_BINARY};
 
 // The connection strings for bootnodes
 const BOOTNODE_1_DNS: &str = "bootnode-tn-1.zkverify.io";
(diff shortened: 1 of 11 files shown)


node/src/service.rs
```diff
@@ -190,7 +196,11 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
         &config.chain_spec,
     );
     let (grandpa_protocol_config, grandpa_notification_service) =
-        sc_consensus_grandpa::grandpa_peers_set_config(grandpa_protocol_name.clone());
+        sc_consensus_grandpa::grandpa_peers_set_config::<_, Network>(
+            grandpa_protocol_name.clone(),
+            NotificationMetrics::new(None),
```
DanieleDiBenedetto, line 201: Shouldn't we attach our Prometheus metrics here ?

runtime/src/lib.rs
```diff
@@ -456,13 +456,8 @@ parameter_types! {
 impl pallet_treasury::Config for Runtime {
     type PalletId = TreasuryPalletId;
     type Currency = Balances;
-    type ApproveOrigin = EitherOfDiverse<EnsureRoot<AccountId>, Treasurer>;
```
DanieleDiBenedetto, line 459: What would be the track in the referenda pallet ? "Treasurer" ?

runtime/src/lib.rs
```diff
@@ -369,7 +369,7 @@ impl_opaque_keys! {
 
 impl pallet_transaction_payment::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
-    type OnChargeTransaction = CurrencyAdapter<Balances, ()>;
+    type OnChargeTransaction = FungibleAdapter<Balances, ()>;
```
DanieleDiBenedetto, line 372: Main differences w.r.t. Currency Adapter ?

node/src/chain_spec.rs
```diff
@@ -152,7 +153,7 @@ pub fn development_config() -> Result<ChainSpec, String> {
 pub fn local_config() -> Result<ChainSpec, String> {
     Ok(ChainSpec::builder(
         WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?,
-        None,
+        Default::default(),
```
DanieleDiBenedetto, line 156: What's the default here ?

runtime/src/lib.rs
DanieleDiBenedetto: Maybe it could be worth adding a test showcasing the new flow ?

node/src/service.rs
rdoria1: Didn't know we had Prometheus metrics, in that case yes, I think they should go here. I'll dig a bit more.

node/src/chain_spec.rs
rdoria1: my bad, this was something I forgot to undo due to some change that was not needed in the end.
Bottom line: No need to do this, I'll undo it.

runtime/src/lib.rs
rdoria1: CurrencyAdapter Implements transaction payment for a pallet implementing the Currency trait whereas FungibleAdapter Implements transaction payment for a pallet implementing the  frame_support::traits::fungible trait.

So really the difference is the deprecation of the Currency trait in favor of the Fungible trait, why?

The reason on this change seems to be because of an issue on balance calculation existing on CurrencyAdapter, specified here [](https://github.com/paritytech/polkadot-sdk/issues/1833)

runtime/src/lib.rs
rdoria1: Actually digging a bit more, I found [this](https://github.com/paritytech/polkadot-sdk/issues/138 ) as well

```The old treasury pallet's dispatchables to create/reject/approve a proposal now replaced by the spend dispatchable.

Deprecated dispatchables:

pallet_treasury.propose_spend
pallet_treasury.reject_proposal
pallet_treasury.approve_proposal
```

The comment on the PR refers to a point time even before removing the ApproveOrigin type 

Also `spend` in the new version is defined as  "Propose and approve a spend of treasury funds, allows spending any asset kind managed by the treasury"

bottom line, yes we could add a new test to call `spend` or `spend_local` and yes I think the track is Treasurer

node/src/service.rs
DanieleDiBenedetto: @drgora

DanieleDiBenedetto:
Closing in favor of #168

PR #163, opened 2024-10-28
Relay Chain

drgora:
This PR supersedes #127.

This PR adds implementations for a relay chain node, and a parachain node. With this PR, the structure of the repository becomes the following:
- mainchain crate: The solo chain `zkv-node`, which should be compatible with the ZKV testnet
- zkv-relay crate: The relay chain `zkv-relay`, which integrates the solo chain node with support for parachains
- paratest crate: The parachain node `paratest-node`, provided as parachain example *for development only*

Instructions to build the solo chain node have been updated in `README.md`. Instructions to build and run the new components are in `RELAY.md`.

The runtime is shared for both chain and the `relay` feature is used to enable the relay chain specific features.


.github/scripts/mount_disk.sh
```diff
@@ -19,27 +19,22 @@
 # cached docker images.
 #
 echo "=============================================================================="
```
drgora, line 21: Forgot to remove this entire file, consider it done already.

DanieleDiBenedetto (approved):
LGTM !!!!!!!!!!!!!

runtime/src/lib.rs
```diff
@@ -842,6 +897,95 @@ construct_runtime!(
...
+
+        // Utility modules.
+        Utility: pallet_utility = 30,
+        Multisig: pallet_multisig = 31,
+        Proxy: pallet_proxy = 32,
+
+
+        // Pallets that we know are to remove in a future. Start indices at 50 to leave room.
+        Sudo: pallet_sudo = 50,
+        ImOnline: pallet_im_online = 51,
```
la10736, line 947: Maybe we can remove it?

runtime/src/paratest_genesis
la10736, line 1: We should not commit it. We should review a bit the migration code to get the file in a `staging` folder that we can add to `.gitignore`.

The review the `RELAY.md`

runtime/src/paratest_wasm
la10736, line 1: See above comment

native/Cargo.toml
```diff
@@ -34,4 +34,5 @@ std = [
     "dep:zksync-era-verifier-deserialize",
     "dep:risc0-verifier",
     "dep:ultraplonk_verifier",
+    "hp-verifiers/std",
```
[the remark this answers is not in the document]
drgora: Good catch, removed the duplicate

runtime/src/paratest_genesis
drgora: Good point, done

pallets/verifiers/Cargo.toml
```diff
@@ -38,7 +38,7 @@ default = ["std"]
 std = [
     "codec/std",
     "frame-support/std",
-    "frame-system/std",
+    "frame-system/default",
```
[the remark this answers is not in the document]
drgora: It looks like we don't, moved back to `std`

primitives/hp-verifiers/Cargo.toml
```diff
@@ -24,4 +24,5 @@ std = [
     "sp-core/std",
     "sp-std/std",
     "sp-io/std",
+    "sp-io/default",
```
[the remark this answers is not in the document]
drgora: It looks like we don't, removed it

runtime/src/lib.rs
drgora: I believe we should really consider this change carefully and see if it aligns well with our use case.
Anyways, for now I'm ok with keeping this aligned with Polkadot, and I removed it.

relay-node/service/chain-specs/zkverify_testnet.json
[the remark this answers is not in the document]
drgora: Yeah I agree, removed

.gitignore
```diff
@@ -11,6 +11,7 @@ zombienet-tests/yarn.lock
 zombienet-tests/bin
 lcov.info
 benchmarking_errors.txt
+staging/
```
la10736: Just a question if you add this here it should not possible to remove it but, the `.gitignore` in `staging` folder that ignore all files instead `.gitignore` should already do the job.

.gitignore
```diff
@@ -21,3 +22,5 @@ coverage_summary.txt
 
 # Ignore external libraries
 deps/
+runtime/src/paratest_genesis
+runtime/src/paratest_wasm
```
la10736: If I saw well you've removed the migration parachain logic...Maybe we can remove this too.

la10736 (approved):
LGTM

PR #42, opened 2024-04-26
Reviewed docker file and gh actions ci

ysibirski:
Final new version of the PR.

One note here. {{secrets.ENV}} stopped working for some reason under bash scripts. I could not find the reason for it, will try to run it on Monday again but {{vars.ENV}} works just fine :shrug:

PR #171, opened 2024-11-07
Update CI-tag-orchestrator for relay chain configs

yurikoinaba:
Build and publish a Docker image for both solo chain and relay chain configurations
- E2E tests run, but Docker images are published regardless of if they pass or fail (for now)
- Docker images are created for both solo chain configuration and relay chain configuration. No changes in tagging to solo chain configured images. For relay chain configured images, tags are appended with `-relay`, i.e. `0.6.0-relay`, `0.6.0-0.8.0-relay`, `latest-relay`
- Upload binaries as `zkv_runtime.solo.compact.compressed.wasm` and `zkv_runtime.relay.compact.compressed.wasm`


.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -120,7 +121,7 @@ jobs:
     strategy:
       matrix: 
         artifact_name:
-        - ${{ needs.build.outputs.artifact_name_solo }}
+        # - ${{ needs.build.outputs.artifact_name_solo }}
```
yurikoinaba, line 124: Will test w relay chain configured image first, and then run both

docker/scripts/bootstrap.sh
```diff
@@ -25,18 +28,36 @@ if [ "$RELAY_FEATURES" ]; then
...
+  ${CARGO} build -p mainchain "${BUILD_PROFILE}"
 
-echo "----------------------------------------------------------"
-echo "Compile test parachain node"
-${CARGO} build -p paratest-node "${BUILD_PROFILE}"
+  echo "----------------------------------------------------------"
+  echo "Compile relay node"
+  ${CARGO} build -p zkv-relay "${BUILD_PROFILE}" "${RELAY_FEATURES}"
+
+  if [ "${SKIP_PARACHAIN}" != "true" ]; then
```
lander86, line 55: I think for the moment we will always skip parachain build.

docker/scripts/bootstrap.sh
```diff
@@ -47,6 +68,8 @@ echo "----------------------------------------------------------"
 echo "Building relay node image"
 "${SCRIPTS}/build-zkv-relay-image-injected.sh"
 
-echo "----------------------------------------------------------"
-echo "Building parachain node image"
-"${SCRIPTS}/build-paratest-image-injected.sh"
+if [ "${SKIP_PARACHAIN}" != "true" ]; then
```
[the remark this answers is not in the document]
yurikoinaba: I wanted to keep it backwards compatible in case this was also used for testing other things

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
     tags:
       - "[0-9]+.[0-9]+.[0-9]+-[0-9]+.[0-9]+.[0-9]+*"
   workflow_dispatch:
+    inputs:
+      git_tag:
+        type: string
```
ysibirski, line 10: What is the reason to take user input for git tag? :thinking: setup_env.sh scrip is still checking if TAG exist in the upstream and derives from PROD or DEV branches.

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+            target/
+          key: build-cache
       
-      - name: Build/Publish Docker image
-        env:
-          RELEASE_BRANCH: release
-          MAINTAINERS_KEYS: ${{ vars.MAINTAINERS_KEYS }}
-          DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
-          DOCKER_HUB_TOKEN: ${{ secrets.DOCKER_HUB_TOKEN }}
+      - name: Verify release build
```
ysibirski, line 63:
```suggestion
      - name: Verify restored build cache artifacts
```

How about this instead if I understand this correctly

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+          "${GITHUB_WORKSPACE}/docker/scripts/bootstrap.sh" --skip-parachain --skip-compile
+        shell: bash
+
+      - name: Set Docker image names as output
+        id: save-image-names
         run: |
-          source "${GITHUB_WORKSPACE}/ci/setup_env.sh"
-          "${GITHUB_WORKSPACE}/ci/docker.sh" --image-artifact zkverify-image
+          echo "docker_image_solo=horizenlabs/zkverify:latest" >> $GITHUB_OUTPUT
+          echo "docker_image_relay=horizenlabs/zkv-relay:latest" >> $GITHUB_OUTPUT
```
ysibirski, line 79: We should move `horizenlabs` into ENV vars better and then refer to this var under the steps

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+          TAR_FILE_RELAY="${{github.workspace}}/${TAR_FILE_NAME_RELAY}.tar"
+
+          echo "tar_file_name_solo=$TAR_FILE_NAME_SOLO" >> $GITHUB_OUTPUT
+          echo "tar_file_solo=$TAR_FILE_SOLO" >> $GITHUB_OUTPUT
+
+          echo "tar_file_name_relay=$TAR_FILE_NAME_RELAY" >> $GITHUB_OUTPUT
+          echo "tar_file_relay=$TAR_FILE_RELAY" >> $GITHUB_OUTPUT
+
+          docker save "${{ steps.save-image-names.outputs.docker_image_solo }}" -o "${TAR_FILE_SOLO}"
+          docker save "${{ steps.save-image-names.outputs.docker_image_relay }}" -o "${TAR_FILE_RELAY}"
```
ysibirski, line 97: Why do we need `save-image-names` step if we can just use the values we are saving directly here:
`docker save horizenlabs/zkverify:latest -o ....` :thinking:

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
         run: |
-          source "${GITHUB_WORKSPACE}/ci/setup_env.sh"
-          "${GITHUB_WORKSPACE}/ci/docker.sh" --image-artifact zkverify-image
+          echo "docker_image_solo=horizenlabs/zkverify:latest" >> $GITHUB_OUTPUT
+          echo "docker_image_relay=horizenlabs/zkv-relay:latest" >> $GITHUB_OUTPUT
+
+      - name: Save the Docker image as a tarball
+        id: create-tar-file
+        run: |           
+          TAR_FILE_NAME_SOLO="zkverify-solo"
```
ysibirski, line 84: I think this can be an ENV var only for a particular scope therefore you do not need to save it under GITHUB_OUTPUT since it is static and not being dynamically defined during the `run:` block :shrug:

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+          TAR_FILE_SOLO="${{github.workspace}}/${TAR_FILE_NAME_SOLO}.tar"
```
ysibirski, line 85: Same for this one

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+
+          TAR_FILE_NAME_RELAY="zkverify-relay"
+          TAR_FILE_RELAY="${{github.workspace}}/${TAR_FILE_NAME_RELAY}.tar"
```
ysibirski, line 88: I think this can be an ENV var only for a particular scope therefore you do not need to save it under GITHUB_OUTPUT since it is static and not being dynamically defined during the run: block 🤷

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+          echo "tar_file_relay=$TAR_FILE_RELAY" >> $GITHUB_OUTPUT
+
+          docker save "${{ steps.save-image-names.outputs.docker_image_solo }}" -o "${TAR_FILE_SOLO}"
+          docker save "${{ steps.save-image-names.outputs.docker_image_relay }}" -o "${TAR_FILE_RELAY}"
+        
+      - name: Upload Docker Image Artifact (solo chain)
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ steps.create-tar-file.outputs.tar_file_name_solo }}
+          path: ${{ steps.create-tar-file.outputs.tar_file_solo }}
```
ysibirski, line 103: There is a chance these can be ENV VARS

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+      CACHING_MODE: "CACHE_RESTORE"
+  
+  build:
     runs-on: ubuntu-latest
-    needs: execute-system-tests # Wait for e2e tests to finish
+    needs: env-setup
+
+    outputs:
+      artifact_name_solo: ${{ steps.create-tar-file.outputs.tar_file_name_solo }}
+      artifact_name_relay: ${{ steps.create-tar-file.outputs.tar_file_name_relay }}
```
ysibirski, line 32: May not be needed if using ENV vars for those values since they are static :shrug: Look at the comments below

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+          overwrite: true
 
-      - name: Upload runtime artifact
+      - name: Upload Docker Image Artifact (relay chain)
         uses: actions/upload-artifact@v4
         with:
-          name: zkv_runtime.compact.compressed.wasm
-          path: ./zkv_runtime.compact.compressed.wasm
+          name: ${{ steps.create-tar-file.outputs.tar_file_name_relay }}
+          path: ${{ steps.create-tar-file.outputs.tar_file_relay }}
```
ysibirski, line 113: Same comment

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+          compression-level: 0
+          overwrite: true
+
+  test-publish:
+    needs: build
+    strategy:
+      matrix: 
+        artifact_name:
+        - ${{ needs.build.outputs.artifact_name_solo }}
+        - ${{ needs.build.outputs.artifact_name_relay }}
```
ysibirski, line 125: Same here

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -73,35 +151,35 @@ jobs:
...
-                      "text": "*Pull Request:* <${{ github.event.pull_request.html_url || github.event.head_commit.url }}|View PR/Commit>"
-                    }
-                  ]
-                }
-              ]
-            }
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+  #     - name: Notify Slack
```
ysibirski, line 154: Still WIP for the commented out lines?

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,54 +5,132 @@ on:
...
+    strategy:
+      matrix: 
+        artifact_name:
+        - ${{ needs.build.outputs.artifact_name_solo }}
+        - ${{ needs.build.outputs.artifact_name_relay }}
+    
+    uses: ./.github/workflows/CI-test-publish.yml
+    with:
+      artifact_name: ${{ matrix.artifact_name }}
+      git_tag: ${{ github.event.inputs.git_tag }}
```
ysibirski, line 130: What if this job was not triggered manually but rather by PUSH ON TAGS? Will this line be still happy :thinking: 
```
github.event.inputs.git_tag
```

.github/workflows/CI-test-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+      DOCKER_HUB_TOKEN:
+        required: true
+
+jobs:
+  execute-e2e-tests:
+    runs-on: ubuntu-latest
+    name: Execute E2E Tests
+    outputs:
+      docker_image_tag: ${{ steps.set-docker-tag.outputs.docker_image_tag }}
+      test_status: ${{ steps.runtest.outputs.test_status }}
```
ysibirski, line 24: Is it me or `runtest` has to be `run-test` = id on line #42?

.github/workflows/CI-test-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+        required: true
+      DOCKER_HUB_TOKEN:
+        required: true
+
+jobs:
+  execute-e2e-tests:
+    runs-on: ubuntu-latest
+    name: Execute E2E Tests
+    outputs:
+      docker_image_tag: ${{ steps.set-docker-tag.outputs.docker_image_tag }}
```
ysibirski, line 23: I do not see `set-docker-tag` step :thinking:

.github/workflows/CI-test-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Download zkVerify Docker image artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ inputs.artifact_name }}
+          path: ./
+
+      - name: Load Docker image
+        run: |
+          image_name="$(docker load -i "${GITHUB_WORKSPACE}/${image_artifact}.tar" | awk '/Loaded image:/ { print $3 }')"
+          log_info "=== Loaded image ${image_name} ==="
```
ysibirski, line 38: log_info is a function, where is it coming from :thinking:

.github/workflows/CI-test-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Download zkVerify Docker image artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ inputs.artifact_name }}
+          path: ./
+
+      - name: Load Docker image
+        run: |
+          image_name="$(docker load -i "${GITHUB_WORKSPACE}/${image_artifact}.tar" | awk '/Loaded image:/ { print $3 }')"
```
ysibirski, line 37: I do not see where `image_artifact` var is defined :thinking:

.github/workflows/CI-test-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          artifact_name: ${{ inputs.artifact_name }}
+          zkverify_docker_tag: latest
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+
+  publish-docker-image:
+    name: Publish Docker Image
+    runs-on: ubuntu-latest
+    # needs: execute-e2e-tests # Wait for e2e tests to finish
+    # if: ${{ needs.execute-e2e-tests.outputs.test_status == 'success' }}
```
ysibirski, line 53: For future imlementation?

ci/docker.sh
```diff
@@ -54,18 +53,15 @@ if [ "${is_a_release}" = "true" ]; then
...
+    log_info "=== Using Docker image artifact from upstream ==="
     image_name="$(docker load -i "${GITHUB_WORKSPACE}/${image_artifact}.tar" | awk '/Loaded image:/ { print $3 }')"
     log_info "=== Loaded image ${image_name} ==="
-    log_info "=== Renaming image ==="
-    docker tag "${image_name}" "${docker_hub_org}/${docker_image_build_name}:${docker_tag_full}"
   else 
-    log_info "=== Building Docker image: ${docker_hub_org}/${docker_image_build_name}:${docker_tag_full} ==="
-    docker build --build-arg PROFILE=production -f "${docker_file_path}" -t "${docker_hub_org}/${docker_image_build_name}:${docker_tag_full}" .
+    log_info "=== No artifact specified ==="
+    exit 1
```
ysibirski, line 64: This should be `fn_die` vs `log_info` which will be in RED color and has exit 1 embedded. I would also add this smth about `--image-artifact` option for a better clarity of the error message

ci/docker.sh
```diff
@@ -83,9 +79,16 @@ if [ -n "${docker_tag_full:-}" ]; then
     publish_tags=("${docker_tag_full}")
   fi
 
+  # Append -relay to tag names for relay chain images
+  if [[ "${image_artifact}" == *"relay"* ]]; then
```
ysibirski, line 83: The idea is to catch ANYTHING that contains `relay` and append -relay to it? That means string like this will match as well `no_relay_here`. Just curious if we need smth more strict or not really?

ci/docker.sh
```diff
@@ -83,9 +79,16 @@ if [ -n "${docker_tag_full:-}" ]; then
...
+    for publish_tag in "${!publish_tags[@]}"; do
+      publish_tags[$publish_tag]="${publish_tags[$publish_tag]}-relay"
```
ysibirski, line 85:
```suggestion
      publish_tags["${publish_tag}"]="${publish_tags["${publish_tag}"]}-relay"
```

docker/scripts/bootstrap.sh
```diff
@@ -25,18 +28,36 @@ if [ "$RELAY_FEATURES" ]; then
...
+        --skip-parachain)
+            SKIP_PARACHAIN=true
+            shift
+            ;;
+        --skip-compile)
+            SKIP_COMPILE=true
+            shift
+            ;;
+    esac
+done
```
ysibirski, line 43: If you want to fix the alignment 4 spaces vs 2

docker/scripts/my_cargo
```diff
@@ -19,7 +19,7 @@ docker run --rm -ti \
     -v "${HOME}/.gitconfig:/.gitconfig" \
     -v /etc/passwd:/etc/passwd:ro \
     -v /etc/group:/etc/group:ro \
-    -v .:/usr/src/node \
+    -v $(pwd):/usr/src/node \
```
ysibirski, line 22:
```suggestion
    -v "$(pwd)":/usr/src/node \
```

ysibirski (commented):
See some comments

.github/workflows/CI-test-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+          RELEASE_BRANCH: release
+          MAINTAINERS_KEYS: ${{ vars.MAINTAINERS_KEYS }}
+          DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
+          DOCKER_HUB_TOKEN: ${{ secrets.DOCKER_HUB_TOKEN }}
+        run: |
+          source "${GITHUB_WORKSPACE}/ci/setup_env.sh"
+          "${GITHUB_WORKSPACE}/ci/docker.sh" --image-artifact ${{ inputs.artifact_name }}
+        shell: bash
+
+      - name: Upload runtime artifact
```
DanieleDiBenedetto, line 78: Is this workflow called after the tag orchestrator one ? That's why you moved it here ?

.github/workflows/CI-test-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+        uses: actions/upload-artifact@v4
```
DanieleDiBenedetto, line 79: Moreover, if the same docker.sh is being called for both solo chain and relay chain image build, we might want to produce both runtime artifacts and differentiate them via a meaningful name (e.g. `zkv_runtime.solo.compact.compressed.wasm` for solo chain and `zkv_runtime.relay.compact.compressed.wasm` for relay-chain)

docker/scripts/bootstrap.sh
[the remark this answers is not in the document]
DanieleDiBenedetto: Is this script used somewhere else in the workflows ? Otherwise I would keep it as it was before since it was meant to be used more for local testing

DanieleDiBenedetto (changes requested):

.github/workflows/CI-tag-orchestrator.yml
yurikoinaba: Yes, to allow for a manual trigger if needed.

It was mostly useful for testing, but I can also see a scenario where you want to rerun it. If a test failed and it gets fixed, for example.

.github/workflows/CI-tag-orchestrator.yml
yurikoinaba: They need to be outputs to be able to be used in a different job 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#passing-values-between-steps-and-jobs-in-a-workflow

.github/workflows/CI-test-publish.yml
yurikoinaba: Yes, good idea.

I kept it here but I wanted to ask where these artifacts are used and it's needed.

.github/workflows/CI-test-publish.yml
yurikoinaba: It's called from the CI-tag-orchestrator now.

There is a matrix that runs build, test, and publish each for solo and relay chain
https://github.com/zkVerify/zkVerify/pull/171/files#diff-f93eb3b00b22939661e1b3f2100a5684cdea4aeb8580b8814d0b2780025d6ef9R27-R35

ci/docker.sh
yurikoinaba: It can be an exact match on ""zkverify-relay"

.github/workflows/CI-tag-orchestrator.yml
yurikoinaba: Good question, it seems to be fine. I triggered this via tag push and didn't get errors on this line

https://github.com/zkVerify/zkVerify/actions/runs/11800656546

.github/workflows/CI-tag-orchestrator.yml
ysibirski: I see why now. Your gitt_tag was in fact "" empty but build-and-test job is defaulting to github_ref..... tag coming from the job if input is empty, smth to keep in mind if this will not step on the way. When you manually INPUT one tag but `build-and-test` job will use a diff tag to build the project

image

```
steps:
      - uses: actions/checkout@v4
        with:
          ref: ${{ github.event.inputs.git_tag || github.ref }}
          fetch-depth: 0
```

.github/workflows/CI-test-publish.yml
yurikoinaba: Will update the naming to what you suggested

.github/workflows/CI-tag-orchestrator.yml
```diff
@@ -5,103 +5,87 @@ on:
...
   RUST_BACKTRACE: 1
 
 jobs:
-  execute-system-tests:
-    name: Execute System Tests
-    uses: HorizenLabs/zkVerify-qa/.github/workflows/CI-e2e-tests.yml@main
+  env-setup:
+    uses: ./.github/workflows/CI-build.yml
+    with:
+      CACHING_MODE: "CACHE_SAVE"
```
yurikoinaba, line 24: It seems to me that the build files should be generated again for this workflow, thus setting it to "CACHE_SAVE"

Or rather, it seems that these folders can be used from the cache
```            
~/.cargo/bin/
~/.cargo/registry/index/
~/.cargo/registry/cache/
~/.cargo/git/db/
```
But the `/target` folder should be regenerated.

Lmk if that is correct @DanieleDiBenedetto @lander86

.github/workflows/CI-test-publish.yml
DanieleDiBenedetto: > Yes, good idea.
> 
> I kept it here but I wanted to ask where these artifacts are used and it's needed.

Yes these are the wasm binaries we use when performing a runtime upgrade

.github/workflows/CI-tag-orchestrator.yml
DanieleDiBenedetto: @la10736

.github/workflows/CI-tag-orchestrator.yml
la10736: I prefer if you remove them instead of comment out: git stand here exactly for doing this job :smile:

.github/workflows/CI-tag-orchestrator.yml
la10736: Anyway I noted that caching mode is required (I cannot understand why... maybe is a wrong requirement)... so you can use `""` 

We should add to backlog if is the case to remove this requirement

.github/workflows/CI-build-test-publish.yml
```diff
@@ -0,0 +1,159 @@
...
+        with:
+          path: |
+            target/
+          key: build-cache
+      
+      - name: Verify restored build cache
+        run: |
+          ls target
+          ls target/release
+
```
la10736, line 63: I don't think you need them to build docker images: please rid them.

.gitignore
```diff
@@ -2,6 +2,7 @@ target/
 **/*.rs.bk
 .idea
 *.iml
+.secrets
```
la10736, line 5: What is the `secrets` about? (just curiosity)

docker/dockerfiles/zkv-relay.Dockerfile
```diff
@@ -36,7 +36,7 @@ ARG PROFILE="release"
 ARG FEATURES=""
 
 ENV BINARY="${BINARY}" \
-  RUN_USER=hl
+  RUN_USER=user
```
la10736, line 39: That's ok.... but, why did you change it? I spent lot of time to catch the best one :smile: 
`hl` stand for horizen labs

docker/scripts/entrypoint.sh
```diff
@@ -54,6 +54,7 @@ get_arg_name_from_env_name() {
     arg_name="${arg_name//_/-}"
     arg_name="${arg_name,,}"
     arg_name=--"${arg_name}"
+    arg_name=$(echo "$arg_name" | sed -r 's/^---+/--/')
```
la10736, line 57: Please, can you explain here why you need this replacement? When `arg_name` already contains some dashes in front of it?

la10736 (changes requested):
Great, just remove 
- Commented code
- Useless steps in the build docker stage
- Take care the note about use and save cache on tag checking

docker/dockerfiles/zkv-relay.Dockerfile
lander86: zkVerify has been moved to zkVerify org and EDS is the one working on the development. I think we should get rid of everything "hl".

docker/scripts/entrypoint.sh
lander86: With this sed command I'm not adding the double hyphens but I'm catching the cases in which we get 3 hyphens and replacing them with just two (or it will error). This happens because the script is not robust and, if the parameters value has an underscore more, it will append more hyphens than expected (this was happening on zkv-relay). 
The script will be more radically updated later one by completely removing the substitution logic to have something more robust.

.github/workflows/CI-tag-orchestrator.yml
yurikoinaba: I think best if this logic is used when the input is defined

.github/workflows/CI-build-test-publish.yml
```diff
@@ -0,0 +1,149 @@
...
+          name: ${{ needs.build.outputs.artifact_name }}
+          path: ./
+
+      - name: Load Docker image and rename
+        run: |
+          IMAGE_NAME="$(docker load -i "${GITHUB_WORKSPACE}/${{ needs.build.outputs.artifact_name }}.tar" | awk '/Loaded image:/ { print $3 }')"
+          echo $(docker images -q "${IMAGE_NAME}" > /dev/null 2>&1 && echo "Image exists..." || echo "Image missing...")
+          echo "Loaded image ${IMAGE_NAME}"
+          TEST_TAG="test"
+          echo "TEST_TAG=${TEST_TAG}" >> $GITHUB_ENV
```
ysibirski, line 101: Why not to use `test` explicitly without the VAR if we are not setting it dynamically or changing via ENV vars :thinking:

.github/workflows/CI-build-test-publish.yml
```diff
@@ -0,0 +1,149 @@
...
+          DOCKER_HUB_TOKEN: ${{ secrets. DOCKER_HUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+          QA_SLACK_WEBHOOK_URL: ${{ secrets.QA_SLACK_WEBHOOK_URL }}
+
+  publish-docker-image:
+    name: Publish Docker Image
+    runs-on: ubuntu-latest
+    needs: build
+    # needs: [ build, execute-e2e-tests ] # Wait for e2e tests to finish
+    # if: ${{ needs.execute-e2e-tests.outputs.test_status == 'success' }}
```
ysibirski, line 120: Still need this?

docker/scripts/entrypoint.sh
ysibirski, line 57:
```suggestion
    arg_name="$(echo "$arg_name" | sed -r 's/^---+/--/')"
```

ysibirski (approved):
LGTM with couple small comments and one suggestion

PR #174, opened 2024-11-08
Entrypoint script impovement

lander86:
Adding a function to eventually download the chainspec from external into the docker entrypoint script.

### DIFF
--- a/docker/dockerfiles/zkv-node.Dockerfile
+++ b/docker/dockerfiles/zkv-node.Dockerfile
@@ -54,8 +54,11 @@ RUN chmod -R a+rx "/usr/local/bin"
 
 RUN apt-get update \
   && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+  aria2 \
+  ca-certificates \
   curl \
-  && useradd -m -u 1000 -U -s /bin/sh -d /${RUN_USER} ${RUN_USER} \
+  jq \
+  && useradd -m -U -s /bin/bash -d "/${RUN_USER}" "${RUN_USER}" \
   && mkdir -p /data /${RUN_USER}/.local/share \
   && chown -R ${RUN_USER}:${RUN_USER} /data /${RUN_USER} \
   && ln -s /data /${RUN_USER}/.local/share \
--- a/docker/dockerfiles/zkv-node.Dockerfile
+++ b/docker/dockerfiles/zkv-node.Dockerfile
@@ -58,7 +58,7 @@ RUN apt-get update \
   ca-certificates \
   curl \
   jq \
-  && useradd -m -U -s /bin/bash -d "/${RUN_USER}" "${RUN_USER}" \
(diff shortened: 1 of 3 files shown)


docker/dockerfiles/zkv-node.Dockerfile
```diff
@@ -54,8 +54,11 @@ RUN chmod -R a+rx "/usr/local/bin"
 
 RUN apt-get update \
   && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+  aria2 \
+  ca-certificates \
   curl \
-  && useradd -m -u 1000 -U -s /bin/sh -d /${RUN_USER} ${RUN_USER} \
+  jq \
+  && useradd -m -U -s /bin/bash -d "/${RUN_USER}" "${RUN_USER}" \
```
ysibirski, line 61:
```suggestion
  && useradd -m -U -s /bin/bash -d /${RUN_USER} ${RUN_USER} \
```

Do not add quotes here OR add them everywhere else then. It may bring more confusion

docker/dockerfiles/zkv-relay.Dockerfile
```diff
@@ -56,8 +56,11 @@ RUN chmod -R a+rx "/usr/local/bin"
 
 RUN apt-get update \
   && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+  aria2 \
```
ysibirski, line 59: Change line ubuntu version from 22 to 24 
```
FROM ubuntu:24.04 as node
```

docker/scripts/entrypoint.sh
```diff
@@ -68,6 +123,7 @@ echo "ZKV_NODE_KEY_FILE=${ZKV_NODE_KEY_FILE}"
 
 ZKV_CONF_BASE_PATH=${ZKV_CONF_BASE_PATH:-}
 ZKV_CONF_CHAIN=${ZKV_CONF_CHAIN:-}
+ZKV_SPEC_FILE_URL="${ZKV_SPEC_FILE_URL:-}"
```
ysibirski, line 126: This name is diff to what I have used in Horizen repo. I think we should be on the same page either this or the other one. I went with `RELAYCHAIN_SPEC_FILE_URL` since it is easier to understand

ysibirski (changes requested):
See some comments

DanieleDiBenedetto (approved):

docker/dockerfiles/zkv-relay.Dockerfile
lander86: For the moment we're going to keep 22.04 ; we will bring ubuntu 24.04 into a later dockerfile and entrypoint refactory.

docker/scripts/entrypoint.sh
lander86: I think it is better to stick with "ZKV" parameter ; let's either change it in horizen repo (ZKV instead of RELAYCHAIN)

ysibirski (approved):
LGTM

PR #175, opened 2024-11-08
Upgrade Ismp Pallets

seunlanlege:
We've updated the ismp pallets to use `anyhow::Error` instead of the internal `ismp::error::Error`. This should allow for custom error type definitions for custom modules now that `anyhow::Error` also supports the no-std `core::error::Error`.


la10736 (approved):
LGTM

Thanks!!!!!!

rdoria1 (approved):

PR #176, opened 2024-11-12
Remove all orchestrator workflow functionality related to slack messaging

rushby:
Aim to reduce noise in slack for all builds.


.github/workflows/CI-orchestrator.yml
```diff
@@ -31,253 +31,31 @@ jobs:
     with:
       CACHING_MODE: "CACHE_RESTORE"
 
-  check-act:
```
DanieleDiBenedetto, line 34: Why we removed this ? was needed only for Slack ?

DanieleDiBenedetto (approved):

yurikoinaba (approved):

.github/workflows/CI-orchestrator.yml
rushby: I thought it was just being used for the slack message logic

DanieleDiBenedetto: @ysibirski

ysibirski: @rushby I think we should still keep it and DO NOT run `set-overall-result` stage if `needs.check-act.outputs.act == 'false'` because I do not think we want to set GitHub status when we run CI/CD locally.

ysibirski (changes requested):
Let's bring back `act` block and if logic so we do not run `set-github-status` block when using act to run CI/CD locally

PR #178, opened 2024-11-13
Introduce deposit for registering verification keys

lgiussan:


runtime/src/lib.rs
```diff
@@ -774,6 +779,34 @@ impl pallet_proxy::Config for Runtime {
...
+        Identity::max_len::<sp_core::H256>() as u32
+    }
+    fn tickets_key_size() -> u32 {
+        Blake2_128Concat::max_len::<(AccountId, sp_core::H256)>() as u32
+    }
+    fn tickets_value_size() -> u32 {
+        VkRegistrationHoldConsideration::max_encoded_len() as u32
+    }
+    parameter_types! {
+        pub VkRegistrationBaseDeposit: Balance = deposit(4, vks_key_size() + tickets_key_size() + tickets_value_size());
```
DanieleDiBenedetto, line 795: Couple of points here:

- Why 4 items ? You are depositing only 2 keys in the maps `vk_hash` in `Vks `and `(account, vk_hash)` in `Tickets`. 
- I'm not even sure you need to make the user pay also for storing the HoldConsideration in `Tickets` (looking around in all pallets this doesn't seem the case because either they don't use the HoldConsideration or because they embed it in the value item in the map)
- I think we can avoid overabstracting here and write explicitly all the constants (they are well known and fixed anyway), maybe just writing a comment is enough

pallets/verifiers/src/lib.rs
```diff
@@ -137,8 +146,27 @@ pub mod pallet {
             + IsType<<Self as frame_system::Config>::RuntimeEvent>;
```
DanieleDiBenedetto, line 146: I was thinking, shall we enforce a MAX_VK_SIZE (s.t. people are not able to register vks bigger than this) ? Perhaps it's safer (also the preimage pallet does that). We should discuss whether it's better this to be in the config of the pallet or an hardcoded constant somewhere

verifiers/fflonk/src/benchmarking.rs
```diff
@@ -72,7 +73,7 @@ mod benchmarks {
     #[benchmark]
     fn register_vk() {
```
DanieleDiBenedetto, line 74: Should we consider having a max_size vk in terms of the weight required to save it in storage ? The difficulty here is that we would need to "bypass" the validate_vk

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
+
+    #[rstest]
+    fn handle_double_registration_by_same_user(mut def_vk: sp_io::TestExternalities) {
+        def_vk.execute_with(|| {
+            let initial_reserved_balance = Balances::reserved_balance(USER_1);
+            assert_ok!(FakeVerifierPallet::register_vk(
+                RuntimeOrigin::signed(USER_1),
+                Box::new(REGISTERED_VK)
+            ));
+            assert_eq!(Balances::reserved_balance(USER_1), initial_reserved_balance);
```
DanieleDiBenedetto, line 193: Don't forget to update here when we will throw an error for an already registered vk from same user

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
+
+    #[rstest]
+    fn handle_double_registration_by_different_users(mut def_vk: sp_io::TestExternalities) {
+        def_vk.execute_with(|| {
+            let initial_reserved_balance = Balances::reserved_balance(USER_2);
+            assert_ok!(FakeVerifierPallet::register_vk(
+                RuntimeOrigin::signed(USER_2),
+                Box::new(REGISTERED_VK)
+            ));
+            assert!(Balances::reserved_balance(USER_2) > initial_reserved_balance);
```
DanieleDiBenedetto, line 181: Shall we check correct increment of the ref_count here as well ?

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
+    ) {
+        def_vk.execute_with(|| {
+            FakeVerifierPallet::register_vk(RuntimeOrigin::signed(USER_2), Box::new(REGISTERED_VK))
                 .unwrap();
-            System::reset_events();
-        });
-        test_ext
+            FakeVerifierPallet::unregister_vk(RuntimeOrigin::signed(USER_1), REGISTERED_VK_HASH)
+                .unwrap();
+            assert!(FakeVerifierPallet::vks(REGISTERED_VK_HASH).is_some());
```
DanieleDiBenedetto, line 221: Redundant, but maybe better to check ref_count was decremented to 1 ?

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
+        })
+    }
+
+    #[rstest]
+    fn release_deposit(mut def_vk: sp_io::TestExternalities) {
+        def_vk.execute_with(|| {
+            assert!(Balances::reserved_balance(USER_1) > 0);
+            FakeVerifierPallet::unregister_vk(RuntimeOrigin::signed(USER_1), REGISTERED_VK_HASH)
+                .unwrap();
+            assert_eq!(Balances::reserved_balance(USER_1), 0)
```
DanieleDiBenedetto, line 231: Not properly checking what you should here. I think you could take a similar approach like you did in the `hold_a_deposit `test. If one wants to be pedantic, you could compute the actual hold amount and check that indeed the user's account has been replenished with that.

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
-    use hex_literal::hex;
+    #[rstest]
+    fn hold_a_deposit(mut test_ext: sp_io::TestExternalities) {
+        test_ext.execute_with(|| {
+            let initial_reserved_balance = Balances::reserved_balance(USER_1);
+            assert_ok!(FakeVerifierPallet::register_vk(
+                RuntimeOrigin::signed(USER_1),
+                Box::new(42)
+            ));
+            assert!(Balances::reserved_balance(USER_1) > initial_reserved_balance);
```
DanieleDiBenedetto, line 149: If one wants to be pedantic, you could compute the actual hold amount and check that indeed the user's account has been emptied by that exact amount.

pallets/verifiers/src/tests.rs
DanieleDiBenedetto: Also check emission of VkRegistered event ?

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
+    #[rstest]
+    fn handle_double_registration_by_same_user(mut def_vk: sp_io::TestExternalities) {
+        def_vk.execute_with(|| {
+            let initial_reserved_balance = Balances::reserved_balance(USER_1);
+            assert_ok!(FakeVerifierPallet::register_vk(
+                RuntimeOrigin::signed(USER_1),
+                Box::new(REGISTERED_VK)
+            ));
+            assert_eq!(Balances::reserved_balance(USER_1), initial_reserved_balance);
+        })
```
DanieleDiBenedetto, line 194: Also check emission of VkRegistered event ?

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
-        "a65dc57cd8f1e436aaa8a8a473005040a4594f5411e0d9c7c5d7f20630217b79"
-    ));
+    #[rstest]
+    fn unregister_a_previously_registered_vk(mut def_vk: sp_io::TestExternalities) {
+        def_vk.execute_with(|| {
+            assert!(FakeVerifierPallet::vks(REGISTERED_VK_HASH).is_some());
+            FakeVerifierPallet::unregister_vk(RuntimeOrigin::signed(USER_1), REGISTERED_VK_HASH)
+                .unwrap();
+            assert!(FakeVerifierPallet::vks(REGISTERED_VK_HASH).is_none());
+        })
```
DanieleDiBenedetto, line 209: Also check emission of VkUnregistered event ?

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
+        def_vk.execute_with(|| {
+            FakeVerifierPallet::register_vk(RuntimeOrigin::signed(USER_2), Box::new(REGISTERED_VK))
                 .unwrap();
-            System::reset_events();
-        });
-        test_ext
+            FakeVerifierPallet::unregister_vk(RuntimeOrigin::signed(USER_1), REGISTERED_VK_HASH)
+                .unwrap();
+            assert!(FakeVerifierPallet::vks(REGISTERED_VK_HASH).is_some());
+        })
```
DanieleDiBenedetto, line 222: Also check emission of VkUnregistered event ?

pallets/verifiers/src/tests.rs
```diff
@@ -95,32 +137,164 @@ mod register_should {
...
+
+        #[rstest]
+        fn on_wrong_origin(mut def_vk: sp_io::TestExternalities) {
+            def_vk.execute_with(|| {
+                assert_noop!(
+                    FakeVerifierPallet::unregister_vk(
+                        RuntimeOrigin::signed(USER_2),
+                        REGISTERED_VK_HASH
+                    ),
+                    RError::VerificationKeyNotFound
```
DanieleDiBenedetto, line 273: Looking at the test I was thinking that maybe better to raise a different Error here ? Something like "UserNotAuthorized" or "BadOrigin" ?

DanieleDiBenedetto (changes requested):

verifiers/fflonk/src/benchmarking.rs
DanieleDiBenedetto: Also one can consider getting fancier with the benches, e.g. register_vk_not_registered, register_vk_already_registered, unregister_vk_only_ref, unregister_vk_many_ref, but maybe is overdoing

pallets/verifiers/src/lib.rs
```diff
@@ -266,17 +308,27 @@ pub mod pallet {
...
+            if !Tickets::<T, I>::contains_key((&account_id, hash)) {
+                let footprint = Footprint::from_encodable(&vk);
+                let ticket = T::Ticket::new(&account_id, footprint)?;
+                Tickets::<T, I>::insert((account_id, hash), ticket);
+                Vks::<T, I>::mutate(hash, |vk_entry| match vk_entry {
+                    Some(VkEntry { ref_count, .. }) => *ref_count += 1,
+                    None => *vk_entry = Some(VkEntry::new(*vk)),
+                })
+            }
             Self::deposit_event(Event::VkRegistered { hash });
```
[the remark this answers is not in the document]
lgiussan: Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a36f6f05a15ae6e8cf37d5f4224ada15324561fb

pallets/verifiers/src/tests.rs
lgiussan: Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a36f6f05a15ae6e8cf37d5f4224ada15324561fb

pallets/verifiers/src/tests.rs
lgiussan: Just changing the returned error would break the test `unregister_should::fail::on_nonexisting_vk`. So, instead in https://github.com/zkVerify/zkVerify/pull/178/commits/45d90e143cbb44855c43da3213fd743b8ebda58e I added the logic for distinguishing the two possible failure reasons (the vk does not exist, or it exists but was not registered by the caller).
The error handling logic introduces another branch in the code, which is not a "subset" of the execution path covered by the current benchmarks. However, the only expensive operation there is a storage read. And the happy path (covered by the benchmark) already performs storage reads and writes, so it should be strictly more expensive. Therefore, if we stick to a worst-case scenario benchmarking, it shouldn't be necessary to add an additional benchmark for the error handling path.

runtime/src/lib.rs
lgiussan: Changed the number of items from 4 to 2 in https://github.com/zkVerify/zkVerify/pull/178/commits/7fbf1c61868af60b81eef761e00d8ea8677a068c. For the other points, decided to keep it as it is, as discussed offline.

pallets/verifiers/src/lib.rs
lgiussan: At the moment every verifier pallet is already imposing a limit on the size of the vk, so it's already impossible to register arbitrarily big vks. If we deem it necessary to introduce a global common limit, I would propose to do so in a different PR.

verifiers/fflonk/src/benchmarking.rs
lgiussan: - For the first comment, please see the point above.
- For the second comment, I agree that it would be possible to benchmark all of those paths. However, this would require to do the following:
  * Since at the moment benchmarks are not "modular", we would need to write those benchmark variants for every single verifier pallet
  * Since it's not possible to distinguish the specific variant from the inputs without actually executing the extrinsic, we would need to use the `DispatchResultWithPostInfo`
Considering that, in normal workflows, errors in these extrinsic would be catched before the extrinsic gets included in a block, I don't think that the effort is really worth it

pallets/verifiers/src/tests.rs
lgiussan: - I think that the `ref_count` is an internal implementation detail that should not be tested
- Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a4eeb34edcd2f58693e2b957c85ec532be4915e0

pallets/verifiers/src/tests.rs
lgiussan: Please see comment above

pallets/verifiers/src/tests.rs
lgiussan: Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a4eeb34edcd2f58693e2b957c85ec532be4915e0

pallets/verifiers/src/tests.rs
lgiussan: Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a4eeb34edcd2f58693e2b957c85ec532be4915e0

pallets/verifiers/src/tests.rs
lgiussan: Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a4eeb34edcd2f58693e2b957c85ec532be4915e0

pallets/verifiers/src/tests.rs
lgiussan: Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a4eeb34edcd2f58693e2b957c85ec532be4915e0

pallets/verifiers/src/tests.rs
lgiussan: Addressed in https://github.com/zkVerify/zkVerify/pull/178/commits/a4eeb34edcd2f58693e2b957c85ec532be4915e0

DanieleDiBenedetto (approved):

pallets/verifiers/src/mock.rs
```diff
@@ -163,15 +178,49 @@ frame_support::construct_runtime!(
...
     type OnProofVerified = OnProofVerifiedMock;
+    type Ticket = HoldConsideration<
+        AccountId,
+        Balances,
+        HoldReasonVkRegistration,
+        LinearStoragePrice<BaseDeposit, PerByteDeposit, Balance>,
+    >;
     type WeightInfo = MockWeightInfo;
+    #[cfg(feature = "runtime-benchmarks")]
+    type Currency = Balances;
```
la10736, line 205: Maybe we can use `Fungible` here instead: `Currency` is deprecated

la10736: Sorry... is in the wrong place

pallets/verifiers/src/lib.rs
```diff
@@ -137,8 +147,27 @@ pub mod pallet {
             + IsType<<Self as frame_system::Config>::RuntimeEvent>;
         /// Proof verified call back
         type OnProofVerified: OnProofVerified;
+        /// A means of providing some cost while data is stored on-chain.
+        type Ticket: Consideration<Self::AccountId>;
         /// Weights
         type WeightInfo: hp_verifiers::WeightInfo<I>;
+        /// Currency used in benchmarks.
+        #[cfg(feature = "runtime-benchmarks")]
+        type Currency: ReservableCurrency<AccountOf<Self>>;
```
la10736, line 156: Maybe we can use `ModifyFungible` (or something like this) here

pallets/verifiers/src/lib.rs
```diff
@@ -266,17 +309,29 @@ pub mod pallet {
...
             let hash = I::vk_hash(&vk);
-            Vks::<T, I>::insert(hash, vk);
+            ensure!(
+                !Tickets::<T, I>::contains_key((&account_id, hash)),
+                Error::<T, I>::VerificationKeyAlreadyRegistered
+            );
+            I::validate_vk(&vk).map_err(Error::<T, I>::from)?;
+            let footprint = Footprint::from_encodable(&vk);
+            let ticket = T::Ticket::new(&account_id, footprint)?;
+            Tickets::<T, I>::insert((account_id, hash), ticket);
```
la10736, line 330: So, we hold the currency like everyone register his own key and we ignore the reference count... Fair

pallets/verifiers/src/lib.rs
```diff
@@ -293,6 +348,35 @@ pub mod pallet {
...
+            let account_id = ensure_signed(origin)?;
+            if let Some(ticket) = Tickets::<T, I>::take((&account_id, vk_hash)) {
+                ticket.drop(&account_id)?;
+                Vks::<T, I>::mutate_exists(vk_hash, |vk_entry| match vk_entry {
+                    Some(v) => {
+                        if v.ref_count > 1 {
+                            v.ref_count -= 1;
+                        } else {
+                            *vk_entry = None;
+                        }
```
la10736, line 368: consider to use always `saturete_sub(1)` instead and the check if `v.ref_count == 0`

pallets/verifiers/src/lib.rs
```diff
@@ -293,6 +348,35 @@ pub mod pallet {
...
+                    }
+                    None => unreachable!(),
+                });
+                Self::deposit_event(Event::VkUnregistered { hash: vk_hash });
```
la10736, line 372: This event is really wired to me You're emitting it also if the ref count is > 0 ?
Or you add also the account id (?!?!?! really awful) otherwise you should emit it iff you really remove it

pallets/verifiers/src/lib.rs
```diff
@@ -266,17 +309,29 @@ pub mod pallet {
...
-            I::validate_vk(&vk).map_err(Error::<T, I>::from)?;
+            let account_id = ensure_signed(origin)?;
             let hash = I::vk_hash(&vk);
-            Vks::<T, I>::insert(hash, vk);
+            ensure!(
+                !Tickets::<T, I>::contains_key((&account_id, hash)),
+                Error::<T, I>::VerificationKeyAlreadyRegistered
+            );
+            I::validate_vk(&vk).map_err(Error::<T, I>::from)?;
+            let footprint = Footprint::from_encodable(&vk);
```
la10736, line 328: IMHO you should also charge the size for ticket storing and account_id

la10736 (approved):
Just some notes... but no blockers

Great Job!!

pallets/verifiers/src/lib.rs
DanieleDiBenedetto: Is this what we want though ? I expect the frontend could wait for some kind of confirmation that the vk was indeed unregistered (even if it was not actually dropped from the storage)

PR #179, opened 2024-11-15
Enable to share the same native cache  on different source base

la10736:
The base idea is to check before if the cache pointed by the current env variable (if any) is valid.

If not we can create the new cache.

### DIFF
--- a/utils/native-cache/src/config.rs
+++ b/utils/native-cache/src/config.rs
@@ -38,7 +38,13 @@ impl Config {
         })
     }
 
-    fn envs(&mut self) -> &mut Table {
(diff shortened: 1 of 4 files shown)


utils/native-cache/src/config.rs
```diff
@@ -79,14 +85,21 @@ impl Config {
     }
 
     /// Add a dependency to the cargo config.toml file.
```
drgora, line 87: This line seems to be outdated now

drgora (approved):
LGTM!

DanieleDiBenedetto (approved):

PR #185, opened 2024-11-25
Some cache improvements

la10736:
1. 4 kind of chache that cache both registry and taget
2. moved all cache instead the codecoverage one on buildjet service that provide 20Gb free but a alimit on a single cache entry of 5Gb (covergare is more than 8)
3. Disabled `CARGO_INCREMENTAL` because is useless on CI 
4. Disable native cache because again are useless on CI

Little note about _3_. Cargo incremental is useful on develop machine because help on skip some partial artifacts compilation on local sources, but cargo use files timestamp to decide if should compile it or not and on CI all owned sorce files will have a timestamp older than the artifacts on the cache ones. Disabling the incremental build we can save 20/30 % in time

To test this PR I've run a cache manager step and then I'll run an orchestrator.

**ATTENTION**
Some notes about this caching strategy. We decided to use a _cache singleton_ because we want to leverage on job parallelism and our cache blobs are really huge and then all new cache element kick out the old ones really fast. **BUT** is not possible to share cache across different branches (reasonable) but we can only use cache generated from parent: that's means the cache created on main can only be used from the branch that are based after the `main/HEAD` tip at the time the cache is computed. To make it explicit imagine the you generate a new branch on Friday and  then a new PR was merged on main, the cache manager run on Sunday and if you try to execute the orchestrator on your new branch Monday you will not use any cache **till you rebase the branch on main**.

In order to mitigate the last described issue we can introduce a check that just check if the cache are present (in the orchestrator) and fail fast is some was missed with an explicit message that ask to rebase your branch on main (we can introduce a flag to override it if needed). In this case we should save money and time.... I hope.

**CACHE IS A MESS**

### DIFF
--- a/.github/workflows/CI-build.yml
+++ b/.github/workflows/CI-build.yml
@@ -29,24 +29,17 @@ jobs:
       - name: Set up deps cache
         run: mkdir deps
 
-      - name: Restore common cache
+      - name: Restore main cache
         if: ${{ !env.ACT && inputs.CACHING_MODE == 'CACHE_RESTORE' }}
-        uses: actions/cache/restore@v3
+        uses: buildjet/cache/restore@v3
         with:
           path: |
             ~/.cargo/bin/
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
-            deps/
-          key: common-cache
-    
-      - name: Restore build cache
-        if: ${{ !env.ACT && inputs.CACHING_MODE == 'CACHE_RESTORE' }}
-        uses: actions/cache/restore@v3
-        with:
-          path: |
             target/
+            deps/
           key: build-cache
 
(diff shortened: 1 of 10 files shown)


DanieleDiBenedetto (approved):
Thank you very much ! So to recap:
- In order to use the cache we need to make sure that our branch is always aligned with main
- How frequent the cache manager job should be ran on main ? Every time a new branch is merged ?

la10736:
> Thank you very much ! So to recap:
> 
> * In order to use the cache we need to make sure that our branch is always aligned with main
> * How frequent the cache manager job should be ran on main ? Every time a new branch is merged ?

Cache manager run 1 time per week (on Sunday), the important thing is that we should have **PR branches based on cache tip** before run orchestrator.

drgora (approved):
Nice! LGTM, thanks!

PR #187, opened 2024-12-02
XCM weights fixup and tests

drgora:
This PR mainly introduces 2 changes:
- the configuration for XCM benchmarks is fixed, resulting in sensible weights generated for XCM teleports
- a zombienet e2e test is added to verify the teleport and remote execution functionalities for XCM

### DIFF
--- a/runtime/src/weights/pallet_xcm.rs
+++ b/runtime/src/weights/pallet_xcm.rs
@@ -71,7 +71,9 @@ impl<T: frame_system::Config> pallet_xcm::WeightInfo for ZKVWeight<T> {
         //  Measured:  `0`
         //  Estimated: `0`
         // Minimum execution time: 18_446_744_073_709_551_000 picoseconds.
-        Weight::from_parts(18_446_744_073_709_551_000, 0)
+        Weight::from_parts(160_940_000, 6196)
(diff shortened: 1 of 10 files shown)


runtime/src/weights/pallet_xcm.rs
```diff
@@ -71,7 +71,9 @@ impl<T: frame_system::Config> pallet_xcm::WeightInfo for ZKVWeight<T> {
         //  Measured:  `0`
         //  Estimated: `0`
         // Minimum execution time: 18_446_744_073_709_551_000 picoseconds.
-        Weight::from_parts(18_446_744_073_709_551_000, 0)
+        Weight::from_parts(160_940_000, 6196)
```
DanieleDiBenedetto, line 74: Was this already executed on the reference machine ?

DanieleDiBenedetto (approved):

runtime/src/weights/pallet_xcm.rs
drgora: No, these were executed locally, and should be considered placeholders.
I believe we will want to run benchmarks on the reference machine once the XCM configuration will be settled for our intended use cases.

la10736 (approved):
WOW!!!

LGTM

PR #188, opened 2024-12-04
Hyperbridge/Ismp integration

rdoria1:


pallets/hyperbridge_aggregations/src/benchmarking.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    /// Return a whitelisted account with enough founds to do anything.
+    pub fn funded_account<T: Config>() -> T::AccountId {
+        let caller: T::AccountId = whitelisted_caller();
+        T::Currency::set_balance(&caller, BalanceOf::<T>::max_value() / 2u32.into());
+        caller
+    }
+}
+
+#[benchmarks]
```
DanieleDiBenedetto, line 49: Having 3 benchmarks is not needed if in your pallet you're specifying to use only the `dispatch_aggregation `one. So you have two options here:
- You build a logic to specify the correct weight to use based on the situation (something similar to what has been done [here](https://github.com/zkVerify/zkVerify/blob/main/verifiers/groth16/src/lib.rs#L96) for instance)
- Considering there isn't much difference between `dispatch_aggregation` and `dispatch_aggregation_large_timeout` and considering that most likely all the dispatch requests will have a fee attached, perhaps we can keep only the `dispatch_aggregation_large_fee` bench

pallets/hyperbridge_aggregations/src/benchmarking.rs
```diff
@@ -0,0 +1,159 @@
...
+            pallet_ismp::Nonce::<T>::get(),
+            initial_nonce + 1,
+            "Nonce should be incremented"
+        );
+    }
+
+    #[benchmark]
+    fn dispatch_aggregation_large_fee() {
+        let caller: T::AccountId = funded_account::<T>();
+        let large_fee = BalanceOf::<T>::max_value() / 4u32.into();
```
DanieleDiBenedetto, line 122: Minor: I don't expect any significant difference in weight depending on the fee amount

pallets/hyperbridge_aggregations/src/benchmarking.rs
```diff
@@ -0,0 +1,159 @@
...
+        };
+
+        let initial_nonce = pallet_ismp::Nonce::<T>::get();
+        let initial_event_count = frame_system::Pallet::<T>::events().len();
+
+        #[extrinsic_call]
+        dispatch_aggregation(RawOrigin::Signed(caller), params);
+
+        // Assertions
+        assert!(
```
DanieleDiBenedetto, line 76: Are you sure ? Digging through the code it seems that at least `pallet_ismp` should emit a `Request` event

runtime/src/tests/availability.rs
```diff
@@ -340,3 +340,22 @@ fn pallet_aggregate() {
         // just checking code builds, hence the pallet is available to the runtime
     });
```
DanieleDiBenedetto, line 341: missing test for availability of the new pallet you introduced + tests ensuring it is using the correct weights

DanieleDiBenedetto (changes requested):

pallets/hyperbridge_aggregations/src/benchmarking.rs
drgora: I agree with this comment, but I believe there is also another option to mention.

If we believe that `timeout` and `fee` have a non-negligible impact on the weight of the extrinsic, which might be represented by some linear regression, they should probably be analyzed appropriately by the benchmarking framework in the valid ranges, using the appropriate parameter types such as `Linear` (for example see [here](https://github.com/zkVerify/zkVerify/blob/230d38bee61ae718870de4be0f1ed296bf7c1b2a/pallets/aggregate/src/benchmarking.rs#L81)).

I would suggest against the option 1 above in this case, since there are no separate code paths which are taken by the different branches of the to-be switch.

rdoria1: The only reason why I included those 2 options is because they are only 2 params from the `dispatch_aggregation` call where the values can theoretically be large. 
In practice though, I would not expect the timeout to be that value (actually from aggregation we might even put this value to 0 since that means messages never expire and that's probably what we want)

bottom line: I'll keep `dispatch_aggregation_large_fee` and remove `dispatch_aggregation_large_timeout `

pallets/hyperbridge_aggregations/src/benchmarking.rs
rdoria1: Based on the results and my previous comment looks like there is some still, which is why I'm keeping the fee benchmark


```
fn dispatch_aggregation() -> Weight {
        // Proof Size summary in bytes:
        //  Measured:  `9`
        //  Estimated: `3474`
        // Minimum execution time: 9_000_000 picoseconds.
        Weight::from_parts(10_000_000, 3474)
            .saturating_add(T::DbWeight::get().reads(2_u64))
            .saturating_add(T::DbWeight::get().writes(2_u64))
    }
    
    fn dispatch_aggregation_large_fee() -> Weight {
        // Proof Size summary in bytes:
        //  Measured:  `9`
        //  Estimated: `3593`
        // Minimum execution time: 31_000_000 picoseconds.
        Weight::from_parts(32_000_000, 3593)
            .saturating_add(T::DbWeight::get().reads(3_u64))
            .saturating_add(T::DbWeight::get().writes(3_u64))
    }
    ```

File: pallets/hyperbridge_aggregations/src/benchmarking.rs (hunk 3 above)
Comment (rdoria1) [reply pallets/hyperbridge_aggregations/src/benchmarking.rs:76]: Yes, that assert is effectively checking that at least one event was emitted, 
checking `frame_system::Pallet::<T>::events().len() > initial_event_count` which is true.

I wanted to exactly check that Request event but since RuntimeEvent is gated behind #[cfg(test)], we can't use it directly in benchmarks, which is why I only tested the number of events.

File: pallets/hyperbridge_aggregations/src/benchmarking.rs (hunk 1 above)
Comment (DanieleDiBenedetto) [reply pallets/hyperbridge_aggregations/src/benchmarking.rs:49]: That's fine but in the pallet you are still using the dispatch_aggregation bench and not the var_fee one. I guess we can substitute it afterwards, but just want to double check your intent here

File: pallets/hyperbridge_aggregations/src/benchmarking.rs [hunk 5]
```diff
@@ -0,0 +1,134 @@
...
+            "Nonce should be incremented"
+        );
+    }
+
+    #[benchmark]
+    fn dispatch_aggregation_var_fee(n: Linear<1, { get_max_benchmark_iterations() }>) {
+        let caller: T::AccountId = funded_account::<T>();
+
+        let max_fee = BalanceOf::<T>::max_value() / 4u32.into();
+        let fee = max_fee.saturating_mul(n.into()) / get_max_benchmark_iterations().into();
```
drgora, line 97: I believe this is not actually doing what you intended: when  `n` grows past `4`, `max_fee` saturates to `BalanceOf::<T>::max_value()` before the division.

Besides, it should be the benchmark framework that uses a given number of iterations, so you should not care about that directly in your code. In other words, I believe that in this code, `n`  should only specify the valid domain for `fee` (e.g. `Linear<0, { max_fee() }>` with `max_fee()` returning `BalanceOf::<T>::max_value() / 4u32.into()`)

DanieleDiBenedetto: Variability is minimal, but we need to depict the worst case here: let's set some timeout and fee (I don't think it's relevant the timeout and fee amount)

drgora: Ok

drgora (approved):
LGTM, thanks

DanieleDiBenedetto:
@rdoria1 CI passes but I was trying to execute the benchmark tests from withing the hyperbrige-aggregators pallet and it fails. This is non blocking for merging this PR but please double check.

DanieleDiBenedetto (approved):

PR #190, opened 2024-12-09
Fix benchmarks and update weights

drgora:
This PR mainly updates the weights computed on our reference machine with the updated relay node.

This PR also updates the reference machine specs, and adds a minor patch to the runtime to work around hardcoded benchmarks for the treasury pallet.


DanieleDiBenedetto (approved):

runtime/src/weights/pallet_preimage.rs
```diff
@@ -85,12 +85,12 @@ impl<T: frame_system::Config> pallet_preimage::WeightInfo for ZKVWeight<T> {
...
-        // Minimum execution time: 15_419_000 picoseconds.
-        Weight::from_parts(65_199_619, 3556)
-            // Standard Error: 6
-            .saturating_add(Weight::from_parts(1_457, 0).saturating_mul(s.into()))
+        // Minimum execution time: 12_273_000 picoseconds.
+        Weight::from_parts(12_584_000, 3556)
+            // Standard Error: 3
+            .saturating_add(Weight::from_parts(2_472, 0).saturating_mul(s.into()))
             .saturating_add(T::DbWeight::get().reads(2_u64))
             .saturating_add(T::DbWeight::get().writes(2_u64))
```
la10736, line 95: Interesting.... why is so lower?

runtime/src/weights/pallet_ultraplonk_verifier.rs
```diff
@@ -69,15 +69,15 @@ impl<T: frame_system::Config> pallet_ultraplonk_verifier::WeightInfo for ZKVWeig
...
     /// Proof: `Poe::FirstInsertionTime` (`max_values`: Some(1), `max_size`: Some(8), added: 503, mode: `MaxEncodedLen`)
     fn submit_proof() -> Weight {
         // Proof Size summary in bytes:
-        //  Measured:  `361`
+        //  Measured:  `416`
         //  Estimated: `177995`
-        // Minimum execution time: 3_677_836_000 picoseconds.
-        Weight::from_parts(3_694_856_000, 177995)
+        // Minimum execution time: 4_759_305_000 picoseconds.
+        Weight::from_parts(4_769_744_000, 177995)
```
la10736, line 80: Strange 25% slower that one month ago...

la10736 (approved):
We should go little bit in deeper because some results seems strange... but later.

LGTM

PR #193, opened 2024-12-12
Fix benchmarks

drgora:
This PR fixes the few benchmarks that broke with the `EXISTENTIAL_DEPOSIT` workaround, now removed.


DanieleDiBenedetto (approved):
Awesome ! Thanks ! Can you provide the same genesis balance to treasury also in the solo-node chainspec ? I don't envision using it anymore, but never say never :)

drgora:
Applied same modifications to the solo chain_spec, and rebased on main.

la10736 (approved):
LGTM

DanieleDiBenedetto (approved):

PR #194, opened 2024-12-13
feat: plonky2 verifier

NikitaMasych:
# What?

This PR adds [plonky2](https://github.com/0xPolygonZero/plonky2) generated proofs verification in zkVerify.

## Other code infrastructure

1. [plonky2-verifier](https://github.com/distributed-lab/plonky2-verifier)
2. [fork of plonky2](https://github.com/distributed-lab/plonky2/tree/stable)

As I understand, you guys might want to move these repositories to your GitHub organization

Fork of plonky2 is needed because:
1. It supports stable rust.
2. It allows building for `thumbv7em-none-eabi` removing `getrandom` crate, which was causing issues.

## Checkpoints
- [x] formatting 
- [x] build 
- [x] tests


verifiers/plonky2/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+mod resources;
+pub(crate) mod verifier_should;
+mod weight;
+
+pub use weight::WeightInfo;
+
+pub const MAX_VK_SIZE: u32 = 262_144; // 256 KB in bytes
+
+pub type Pubs = Vec<u8>;
+pub type Proof = Vec<u8>;
```
DanieleDiBenedetto, line 23: Why not enforcing a maximum size also on Pubs and Proofs ?

DanieleDiBenedetto (changes requested):

verifiers/plonky2/src/lib.rs
NikitaMasych: Primarily because `Vk` must be limited, since it is constrained by traits we need it to implement. That is the reason we use `BoundedVec` instead of `Vec` as in other types. I assume the idea is to allow a bigger range of proofs users can verify. Maybe you know whether 256 KB is sufficient?

verifiers/plonky2/Cargo.toml
```diff
@@ -0,0 +1,57 @@
+[package]
+name = "pallet-plonky2-verifier"
+version = "0.1.0"
+description = "A plonky2 verifier pallet implementation based on pallet-verifiers abstraction."
+homepage.workspace = true
+edition.workspace = true
+authors.workspace = true
+repository.workspace = true
+license = "GPL-3.0"
```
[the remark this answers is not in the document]
NikitaMasych: That's just after-templating residue, I'll adjust this one. What licence you think would be better here?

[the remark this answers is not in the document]
DanieleDiBenedetto: Apache-2 would be fine, thanks :)

verifiers/plonky2/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn hash_context_data() -> &'static [u8] {
+        b"plonky2"
+    }
+
+    fn verify_proof(
+        vk: &Self::Vk,
+        raw_proof: &Self::Proof,
+        raw_pubs: &Self::Pubs,
+    ) -> Result<(), VerifyError> {
+        verify_default_poseidon(vk, raw_proof, raw_pubs)
```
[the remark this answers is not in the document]
NikitaMasych: Sure, I'll see that I can do

verifiers/plonky2/src/lib.rs
DanieleDiBenedetto: It would be wise to put a limit on proof and pubs as well. I think 256kB might be fine, perhaps we can expose this as configuration parameter ?

runtime/src/lib.rs
```diff
@@ -1026,6 +1026,29 @@ impl IsmpRouter for ModuleRouter {
     }
 }
 
+parameter_types! {
+    pub const Plonky2MaxPubsSize: u32 = 262_144;
```
DanieleDiBenedetto, line 1030: Minor: Perhaps we can reduce this, don't envision applications with that many public inputs

DanieleDiBenedetto (approved):
LGTM !
One more request on my side considering you have implemented the custom gate serializer as well as the possibility to specify the config (poseidon/keccak) in the vk.
We need a simple way for people using plonky2 to serialize/deserialize artifacts using the custom gate serializer and the configuration options, in a format that will be accepted by zkVerify, in an easy and dev-friendly way.
I was thinking you could provide utilities/examples in the plonky2-verifier repo.
A tutorial on our [documentation ](https://github.com/HorizenLabs/zkverify-docs) would be useful as well.
Many thanks !

runtime/src/lib.rs
NikitaMasych: Clarification: these are viewed here as bytes, not quantitive amount. Do you have in mind appropriate value we can set?

NikitaMasych:
Added documentation:

https://github.com/zkVerify/zkverify-docs/pull/100

verifiers/plonky2/src/benchmarking.rs
```diff
@@ -0,0 +1,213 @@
...
+
+fn init<T: pallet_aggregate::Config>() -> (T::AccountId, u32) {
+    let caller: T::AccountId = funded_account::<T>();
+    let domain_id = 1;
+    insert_domain::<T>(domain_id, caller.clone(), Some(1));
+    (caller, domain_id)
+}
+
+#[benchmarks(where T: pallet_verifiers::Config<Plonky2<T>> + pallet_aggregate::Config)]
+mod benchmarks {
```
DanieleDiBenedetto, line 26: Is there any relevant difference between the Poseidon and Keccak configuration that might be worth to catch in the benchmarks ?

NikitaMasych: If you find it desirable to add, no problem 🙂

DanieleDiBenedetto: If you bench locally and see some non-negligible differences, it might be worth to add them, thanks !

DanieleDiBenedetto: Also, which proof did you use to run the bench ?
I tried to run benches on our reference machine and it only take 11ms which seems to me a little bit too fast.
Are we able to take a more complex proof from somewhere and use that for benches ?

NikitaMasych: I've using this circuit from `plonky2` repository examples:

https://github.com/NikitaMasych/fibonacci-nostd

DanieleDiBenedetto: I'm afraid we should find a more complex real-world example, can you try browsing around a bit to see if you find something ? Thanks !

verifiers/plonky2/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+sp-core = { workspace = true }
+hp-verifiers = { workspace = true }
+log = "0.4.20"
+hex-literal = { workspace = true, optional = true }
+frame-system = { workspace = true }
+frame-benchmarking = { workspace = true, optional = true }
+pallet-verifiers = { workspace = true }
+sp-io = { workspace = true, optional = true }
+frame-support = { workspace = true }
+plonky2-verifier = {git = "https://github.com/distributed-lab/plonky2-verifier", default-features = false}
```
DanieleDiBenedetto, line 26: Can we create a tag and we reference it ? To avoid unwanted updates

NikitaMasych: Sure we can

NikitaMasych: Set this to `0.1.0`

NikitaMasych:
Rebased with respect to new changes from main.

Run `cargo fmt` and `cargo clippy` on `verifiers/plonky2` and fixed some minor issues.

Issued `v0.1.0` tag on `plonky2-verifier`.

NikitaMasych:
My apologies for the delay with running ci locally, I will be busy a few more days, afterwards I would be able to resolve pending things

verifiers/plonky2/src/benchmarking.rs
NikitaMasych: I've searched for some more complex `plonky2` circuit, and here are some considerations:

plonky2 is a type of zkp, which has poly-logarithmic asymptotics on verification, meaning that even for really-really huge circuit, verification speed might increase only negligibly. I've proved it practically, changing my circuit to compute not 100-th element of fibonacci sequence, but rather a 100-million'th - that required me spending appr. 16 minutes on M2 pro with 32 GB ram to prove, but the verification speed changed from 1.1147 ms to 4.7470 ms

NikitaMasych: So, you probably shall not worry that results from small circuit verification times might not be really showcasing plonky2-verifier efficiency

NikitaMasych: I can replace currently used proof in tests/benches in zkVerify-plonky2 to mentioned above with 100-millionth fibonacci though

DanieleDiBenedetto:
> My apologies for the delay with running ci locally, I will be busy a few more days, afterwards I would be able to resolve pending things

No worries, thank you very much ! Question: would we be able to also verify recursive proofs with this verifier ?

runtime/src/lib.rs
```diff
@@ -1040,6 +1040,29 @@ impl IsmpRouter for ModuleRouter {
...
+    pub const Plonky2MaxProofSize: u32 = 262_144;
```
DanieleDiBenedetto, line 1045: Looking at Plonky2 codebase it seems it's possible to compress the proof. You save around 10 KB of data which might be enough for us to consider

runtime/src/lib.rs
```diff
@@ -1040,6 +1040,29 @@ impl IsmpRouter for ModuleRouter {
...
+    pub const Plonky2MaxVkSize: u32 = 262_144;
```
DanieleDiBenedetto, line 1046: This might be too much indeed, I think we can set it to something like 50_000 and be happy with it

runtime/src/lib.rs
DanieleDiBenedetto: Indeed we can set this to a much smaller value, like a size equivalent to 64 public inputs if we want to be really safe

DanieleDiBenedetto:
> > Question: would we be able to also verify recursive proofs with this verifier ?
> 
> I don't see why not, however, I would check it practically 🙂

Thanks much ! Just to be fully sure, please check we are also able to verify proofs using lookups :)

verifiers/plonky2/src/benchmarking.rs
DanieleDiBenedetto: So verifier complexity increase with the degree of the circuit (e.g. num gates)
I was running this bench https://github.com/0xPolygonZero/plonky2/blob/main/plonky2/examples/bench_recursion.rs on my machine and it takes 4-5 ms to verify a proof of degree 2^12 - 2^14 going up to 7-8 ms for a proof of degree 2^19-2^21.
If we want to be over-pedantic, one might want to consider having different benches for different proof degrees (which can be read from `vk.common.degree()`) or just set it to a very big number (like a proof with degree 2^25) and over-estimate the weight (excess fee is reimbursed anyway)

lgiussan (changes requested):
Thanks for the contribution!

Apart from the few minor comments I left, the biggest issue I see is about weights.
The verifier currently supports arbitrary configurations, and different configurations bring, a priori, different running times for the verification process. 
However, at the moment, benchmarks do not capture this variability: the weight of the `verify_proof` extrinsic is fixed, and does not reflect the worst-case scenario (the proof, vk, and pubs used in the benchmark are all smaller than their maximum allowed size).
This could open the door to DoS attacks to the chain.

In order to tackle this we would need to:
1. Understand how configuration parameters impact the running time of the verification algorithm, ideally by building a parametric model of the running time as a function of those parameters.
2. Leveraging this knowledge, find out which is the worst-case configuration compatible with the limitations we impose (currently only on proof, vk, and pubs size, but we could place additional, more fine-grained limitations on configuration parameters if needed).
3. Run the benchmarks using an instance reflecting that worst-case configuration. This would protect the chain from DoS attacks, but would make users pay more than necessary.
4. (Optional, can be done also at a later time) Refine the benchmarks, so that they span the entire configuration space. Using those results, implement a refund mechanism which refunds the appropriate amount to users submitting proof instances smaller than the worst-case one

lgiussan:
> 
> In order to tackle this we would need to:
> 
>     1. Understand how configuration parameters impact the running time of the verification algorithm, ideally by building a parametric model of the running time as a function of those parameters.
> 
>     2. Leveraging this knowledge, find out which is the worst-case configuration compatible with the limitations we impose (currently only on proof, vk, and pubs size, but we could place additional, more fine-grained limitations on configuration parameters if needed).
> 
>     3. Run the benchmarks using an instance reflecting that worst-case configuration. This would protect the chain from DoS attacks, but would make users pay more than necessary.
> 
>     4. (Optional, can be done also at a later time) Refine the benchmarks, so that they span the entire configuration space. Using those results, implement a refund mechanism which refunds the appropriate amount to users submitting proof instances smaller than the worst-case one.
> 

Another, easier, possibility would be to restrict the configuration instead, e.g. to the [standard one](https://github.com/0xPolygonZero/plonky2/blob/main/plonky2/src/plonk/circuit_data.rs#L100-L119). This would somewhat limit the applicability of our verifier, but I think it's an acceptable tradeoff for quickly releasing a working and safe plonky2 verifier, and start gathering data about its usage.

verifiers/plonky2/src/benchmarking.rs
NikitaMasych: I've restricted config to standard one by advice from @lgiussan: https://github.com/zkVerify/zkVerify/pull/194#issuecomment-2654091196.

Hence, we will have a fixed degree.

NikitaMasych:
> Another, easier, possibility would be to restrict the configuration instead, e.g. to the [standard one](https://github.com/0xPolygonZero/plonky2/blob/main/plonky2/src/plonk/circuit_data.rs#L100-L119). This would somewhat limit the applicability of our verifier, but I think it's an acceptable tradeoff for quickly releasing a working and safe plonky2 verifier, and start gathering data about its usage.

Done in:

https://github.com/distributed-lab/plonky2-verifier/commit/e8160ebadc87ebfc5e1638e2d084b9ffe8661b69

verifiers/plonky2/src/benchmarking.rs
DanieleDiBenedetto: @NikitaMasych actually no, the degree depends on the number of gates in the circuit, regardless of the configuration

NikitaMasych: Ah, my bad, I've confused `FriParams` with `FriConfig`

NikitaMasych: I've set circuit we are using in resources to one with 2^20 degree (got it in [bench_recursion.rs](https://github.com/0xPolygonZero/plonky2/blob/main/plonky2/examples/bench_recursion.rs) - inner one, setting `size` to 20.

runtime/src/lib.rs
NikitaMasych: I've brought in custom `Proof` struct to support this; you can check it out here:

https://github.com/distributed-lab/plonky2-verifier/commit/7b9efd3f68dda7fa0bb7746ee72576f7fe08b788

PR #197, opened 2024-12-18
Avoid duplicated vk checks

lgiussan:
Currently, some verifiers perform some expensive checks when loading a registered vk from storage, even if these checks were already performed during vk registration.
This PR avoids duplication of the most expensive of such checks (namely that the EC points lie on curve) for the groth16, and fflonk verifiers.
No modifications were made to the other verifiers for the following reasons:
- Risc0 vk is an arbitrary U256, and no checks are performed
- ZkSync doesn't have a vk
- ProofOfSQL does not perform any check
- Ultraplonk will be soon completely replaced by a wasm verifier mirroring Solidity implementation


DanieleDiBenedetto (approved):

la10736 (approved):
LGTM

PR #199, opened 2024-12-19
Re-enable zombienet test for Horizen registration

drgora:
This PR updates the Horizen test chainspec (and id), so that the related zombienet can be re-enabled.


DanieleDiBenedetto (approved):
LGTM ! Couple of comments here:
- Would be nice to find a way to always parse the latest chainspec from Horizen2 (e.g. latest tag or latest commit from main branch)
- The CI will fail if this test fails. I'm not sure that should be the expected behaviour (we need to do the changes that we need, and H2 parachain should adapt to us). Perhaps throwing a warning instead of failing the CI for this specific test would be better ?

cc. @paolocappelletti

la10736 (approved):
LGTM

PR #201, opened 2024-12-23
Claim Pallet

DanieleDiBenedetto:
This PR introduces a pallet implementing the possibility of making airdrops on-chain and letting **Beneficiaries**
manually claim the amount of tokens they have right to.

- Only **ManagerOrigin** is able to start and end airdrops, as well as adding and removing beneficiaries with their rightful balance.
- Currently it possible only to held one airdrop at a time.
- When airdrop ends, all the funds still available in the pallet's associated account are transferred to **UnclaimedDestination**.
- It is required that the pallet's associated accounts always holds enough funds for the Beneficiaries. Adding Beneficiaries without having enough funds will cause an error.
- All the extrinsics don't charge any fee to the caller


pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+                    } else {
+                        // Account doesn't exist. Add its token amount to the required amount this pallet's account should have
+                        required_amount = required_amount.saturating_add(*amount);
+                        log::trace!("Added beneficiary {account:?}. Can claim: {amount:?}");
+                    }
+
+                    // Cannot cover for all the tokens, raise an error
+                    if required_amount > available_amount {
+                        Err(Error::<T>::NotEnoughFunds)?;
+                    }
```
drgora, line 287: If I'm understanding the logic correctly, I believe this check should be done out of the loop, as it may temporarily fail (while it shouldn't) during the iterations.

Also, I'm wondering whether it would make sense to make this entire function atomic: either it succeeds and `Beneficiaries` is updated, or it fails and the state is not changed at all.

drgora (changes requested):
I'm not sure that allowing the execution of `do_add_beneficiaries` while the airdrop is active is safe. The expected interaction is not clear, and a different order of the extrinsics might result in unexpected and unwanted behaviors.

For example, consider the following scenario:
```
add_beneficiaries(addr1, 1000);
claim_for(addr1);
add_beneficiaries(addr1, 2000);
```
This might even be a legit scenario, but what if the second call to `add_beneficiaries` was only meant to update the airdrop for `addr1` from `1000` to `2000`, instead of assigning a total of `3000`?
This dependency on the order of the extrinsics is undesirable. I guess we might consider adding a `paused` state for the airdrop, and restrict any beneficiary manipulation to be executed only in a paused state? Or else, we could forbid updating existing entries in the beneficiary list altogether?

pallets/claim/src/lib.rs
DanieleDiBenedetto: This is a good point indeed, but honestly I don't envision any of the add/remove beneficiaries being ever called after having added beneficiaries the first time. On top of this, considering they could only be called by a Super User or Treasury, we can safely assume that they won't be that frequent/superficial to cause this kind of issues.
I already felt overdesigning with these mechanisms, and thus I don't feel complicating the logic furthermore for something that will likely never happen.
But I'm happy to discuss further you think it's indeed worth the effort :)

DanieleDiBenedetto: Extrinsic execution should already be atomic: if it fails in the middle of the execution, the state will stay untouched

drgora: Right, the atomicity is guaranteed by the framework, makes sense. But I still believe the first point should be addressed: in the current state, the update might fail even if it is correct.
For example, imagine having a pot of `10`, distributed like `(addr1, 5), (addr2, 5)`; updating that to `(addr1, 6), (addr2, 4)` might fail with `NotEnoughFunds`, depending on the order of the beneficiaries map.

drgora:
> @drgora This is a good point indeed, but honestly I don't envision any of the add/remove beneficiaries being ever called after having added beneficiaries the first time.

I agree, so maybe we can just remove it?

> On top of this, considering they could only be called by a Super User or Treasury, we can safely assume that they won't be that frequent/superficial to cause this kind of issues.

I believe this is not an issue of being superficial, and might be actually open room for attacks: a malicious (or just lucky) beneficiary, might be able to observe the `add_beneficiaries` extrinsic in the mempool, before it is added to a block. If she manages to front-run that with her `claim` extrinsic, she might be able to obtain more tokens than intended.

pallets/claim/src/lib.rs
DanieleDiBenedetto: Ah right, good catch !

drgora:
>     * Remove the "remove_beneficiaries" and keep "add_beneficiaries" but it won't be possible to modify already existing beneficiaries' amounts (I could indeed see a situation in which we might want to add a new beneficiaries and don't want to wait for the next airdrop. Well, we could introduce the possibility of managing multiple airdrops but again, at the moment, it feels like overdesign)

I totally agree on this, forbidding any modification to existing allocations (besides zeroing them) seems like a simple yet effective solution.

pallets/claim/src/benchmarking.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        // Prepare beneficiaries and sufficient amount
+        let (beneficiaries, total_amount) = get_beneficiaries_map::<T>(n);
+        let _ = T::Currency::mint_into(
+            &Pallet::<T>::account_id(),
+            total_amount.saturating_mul(2u32.into()), // Just to be extra safe
+        )
+        .unwrap();
+
+        // Worst case scenario: all the amounts of existing beneficiaries have been modified
```
drgora, line 118: It looks like this comment can be removed?

drgora (approved):
LGTM, I just left few comments here and there, none of which should be a blocker

pallets/claim/src/benchmarking.rs
```diff
@@ -0,0 +1,134 @@
+// Copyright 2024, Horizen Labs, Inc.
```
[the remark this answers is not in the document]
DanieleDiBenedetto: @lgiussan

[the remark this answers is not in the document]
lgiussan: If it's doable, I would sugegst yes

pallets/claim/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+version = "0.1.0"
+description = "A pallet to claim airdropped tokens"
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+repository.workspace = true
+license = "Apache-2.0"
+
+[dependencies]
+log = "0.4.20"
```
la10736, line 12: Maybe could be a a nice idea to start to move all crates dependency (at least the common ones) in the workspace Cargo.toml.

What do you think?

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+        type WeightInfo: WeightInfo;
+
+        /// The maximum number of beneficiaries used in benchmarks.
+        #[cfg(feature = "runtime-benchmarks")]
+        const MAX_BENEFICIARIES: u32;
+    }
+
+    /// Candidates eligible to receive an airdrop with the associated balance they have right to
+    #[pallet::storage]
+    #[pallet::getter(fn beneficiaries)]
```
la10736, line 102: You can remove all getter: are unused

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+
+        /// Manager allowed to begin/end airdrops and add beneficiaries
+        type ManagerOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+
+        /// Type for processing spends of [Self::AssetKind] in favor of [`Self::Beneficiary`].
+        type Paymaster: Pay<
+            Beneficiary = Self::AccountId,
+            AssetKind = (),
+            Balance = BalanceOf<Self>,
+        >;
```
la10736, line 84: Maybe you don't need this abstraction at all. Indeed, it's nice, but you cannot use this abstraction to inspect the account and to move founds in. Moreove this can be not coherent with the pallet `account_id` used to check the ballance and move the unclaimed founds.

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+        /// Attempt to perform an action implying an open airdrop, while it has already ended
+        AlreadyEnded,
+    }
+
+    impl<T: Config> Pallet<T> {
+        /// The account ID of the claim pot.
+        ///
+        /// This actually does computation. If you need to keep using it, then make sure you cache the
+        /// value and only call this once.
+        pub fn account_id() -> T::AccountId {
```
la10736, line 205: According to the comment save it in a storage value to not recompute it every time ... or remove the comment.

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+                        // Determine who is the beneficiary
+                        let beneficiary = beneficiary.unwrap_or(origin);
+                        // Execute payment
+                        let available = Self::pot();
+                        if *amount > available {
+                            log::warn!("Claimable amount {amount:?} bigger than total available {available:?}");
+                            Err(Error::<T>::PayoutError)?; // Prevent going under the existential deposit of the account
+                        }
+                        let payment_id = T::Paymaster::pay(&beneficiary, (), *amount)
+                            .map_err(|_| Error::<T>::PayoutError)?;
```
la10736, line 234: Again, maybe is better to remove `Paymaster` abstraction because here you didn't have any assurance that the Paymaster have enough founds.

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+                        // Subtract the claimed token from the TotalClaimable
+                        TotalClaimable::<T>::mutate(|required_amount| {
+                            *required_amount = required_amount.saturating_sub(*amount)
```
la10736, line 237: Use the `defensive` variation instead

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+            /// The id of the airdrop that has just ended
+            airdrop_id: u64,
+        },
+    }
+
+    /// Error for the claim pallet.
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Attemp to start a new airdrop while there is one already in progress
+        AlreadyStarted,
```
la10736, line 185:
```suggestion
        /// Attempt to start a new airdrop while there is one already in progress
```

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+                *amount = match amount {
+                    // Account is eligible to get an airdrop
+                    Some(amount) => {
+                        // Determine who is the beneficiary
+                        let beneficiary = beneficiary.unwrap_or(origin);
+                        // Execute payment
+                        let available = Self::pot();
+                        if *amount > available {
+                            log::warn!("Claimable amount {amount:?} bigger than total available {available:?}");
+                            Err(Error::<T>::PayoutError)?; // Prevent going under the existential deposit of the account
```
la10736, line 231: Here you can use the substrate's `TokenError:FundsUnavailable` error instead

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+                        // Account already exists
+                        log::warn!("Beneficiary {account:?} already added.");
+                        Err(Error::<T>::AlreadyPresent)?;
+                    } else if amount.is_zero() {
+                        // Attempting to add a beneficiary with nothing to claim
+                        log::warn!("Beneficiary {account:?} with nothing to claim.");
+                        Err(Error::<T>::NothingToClaim)?;
+                    } else {
+                        // Account doesn't exist. Add its token amount to the required amount this pallet's account should have
+                        required_amount = required_amount.saturating_add(*amount);
```
la10736, line 275: again, use defensive variation instead

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+        ) -> DispatchResult {
+            // Check that the pot has enough funds to cover for all the beneficiaries
+            let available_amount = Self::pot();
+            let mut required_amount = TotalClaimable::<T>::get();
+
+            beneficiaries
+                .iter()
+                .try_for_each::<_, DispatchResult>(|(account, amount)| {
+                    if let Some(_) = Beneficiaries::<T>::get(account.clone()) {
+                        // Account already exists
```
la10736, line 266:
```suggestion
                    if Beneficiaries::<T>::contains_key(account) {
```

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+                        log::warn!("Beneficiary {account:?} already added.");
+                        Err(Error::<T>::AlreadyPresent)?;
+                    } else if amount.is_zero() {
+                        // Attempting to add a beneficiary with nothing to claim
+                        log::warn!("Beneficiary {account:?} with nothing to claim.");
```
la10736, line 271: Use guard approach instead:

```rust
if Beneficiaries::<T>::contains_key(account) {
    // Account already exists
    log::warn!("Beneficiary {account:?} already added.");
    Err(Error::<T>::AlreadyPresent)?;
} 
if amount.is_zero() {
    // Attempting to add a beneficiary with nothing to claim
    log::warn!("Beneficiary {account:?} with nothing to claim.");
    Err(Error::<T>::NothingToClaim)?;
}
// Account doesn't exist. Add its token amount to the required amount this pallet's account should have
required_amount = required_amount.saturating_add(*amount);
log::trace!("Added beneficiary {account:?}. Can claim: {amount:?}");
```

It's not _structured_ programming but is nicer and more clear to understand: the business logic is in the main branch code

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+        /// This is an atomic operation. If there isn't enough balance to cover for all the beneficiaries, then none will be added.
+        /// Origin must be the ManagerOrigin.
+        #[pallet::call_index(0)]
+        #[pallet::weight(match &beneficiaries {
+            Some(beneficiaries) => T::WeightInfo::begin_airdrop_with_beneficiaries(beneficiaries.len() as u32),
+            None => T::WeightInfo::begin_airdrop_empty_beneficiaries(),
+        })]
+        pub fn begin_airdrop(
+            origin: OriginFor<T>,
+            beneficiaries: Option<BTreeMap<T::AccountId, BalanceOf<T>>>,
```
la10736, line 318: I'm proposing to remove the option here because I cannot see any advantage on call this function with `None` instead of using an empty map. If you're _scared_ by the `do_add_beneficiaries()` call you can put a guard at the `do_add_beneficiaries()` first line to exit if the map is empty and save two storage read.

This change also should simplify the benchmarks.

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+            }
+
+            // Increase airdrop id
+            AirdropId::<T>::mutate(|id| {
+                let airdrop_id = id.map_or(0, |v| v + 1);
+                *id = Some(airdrop_id);
+                Self::deposit_event(Event::<T>::AirdropStarted { airdrop_id });
+            });
+
+            Ok(PostDispatchInfo {
```
la10736, line 344:
```
Ok(Pays::No.into())
```
Do the job

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+                actual_weight: None,
+                pays_fee: Pays::No,
+            })
+        }
+    }
+}
+
+/// TypedGet implementation to get the AccountId of the Claim pallet.
+pub struct ClaimAccountId<R>(PhantomData<R>);
```
la10736, line 458: You didn't test this public struct

pallets/claim/src/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+        );
+        assert_eq!(TotalClaimable::<Test>::get(), SUFFICIENT_GENESIS_BALANCE);
+        assert!(AirdropActive::<Test>::get());
+        assert_eq!(AirdropId::<Test>::get(), Some(0));
+    })
+}
+
+#[test]
+fn new_aidrop_insufficient_funds() {
+    test_with_configs(
```
la10736, line 135:
```suggestion
fn new_airdrop_insufficient_funds() {
```

pallets/claim/src/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+}
+
+#[test]
+fn claim_wrong_beneficiary() {
+    test_with_configs(
+        WithGenesisBeneficiaries::Yes,
+        GenesisClaimBalance::Sufficient,
+    )
+    .execute_with(|| {
+        assert_err!(
```
la10736, line 256: `assert_nop!` checks that you have an error **and** the storage not change: that's enough to check everything.

pallets/claim/src/tests.rs
la10736, line 1: You have a great coverage (91%) but I saw some issue here

1. You didn't test the _No pay_ feature anymore that is a business requirement
2. Too much assert in tests: ideally a (unit) tests should have just one assert per test.
3. In the negative tests you can leverage on `assert_nop` that also checks that the storage doesn't change. I put a comment in one of them, but is quite general.
4.

pallets/claim/src/benchmarking.rs
```diff
@@ -0,0 +1,133 @@
...
+    beneficiaries
+}
+
+#[benchmarks]
+mod benchmarks {
+
+    use super::*;
+
+    #[benchmark]
+    fn begin_airdrop_empty_beneficiaries() {
```
la10736, line 62: You can remove it

la10736 (changes requested):
Just some minor changes (no critical) and almost about the style and not the logic.

pallets/claim/src/benchmarking.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: will do in another PR, thanks !

pallets/claim/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+                    *is_active = false;
+                    Ok::<_, DispatchError>(())
+                }
+            })?;
+
+            // Remove all beneficiaries entries
+            let _ = Beneficiaries::<T>::clear(u32::MAX, None);
+
+            // Deal with any remaining balance in the pallet's account
+            // TODO: Shall we want to withdraw all the balance in the pallet's account or only up to
```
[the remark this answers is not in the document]
DanieleDiBenedetto: @la10736 any opinion on this ?

pallets/claim/src/lib.rs
la10736: define
```rust
    #[pallet::storage]
    pub type AccountId<T: Config> = StorageValue<_, T::AccountId>;
```
now you can use
```rust
    AccountId::<T>::mutate(|v| {
        let id = v.unwrap_or_else(|| T::PalletId::get().into_account_truncating());
        if v.is_none() {
            *v = Some(id.clone());
        }
        id
    })
```
I don't know if there is a simpler way to do it... If I'm not remember wrong I saw it implemented in a better way in pallet-timestamp or pallet-autority

la10736: I took a look to `pallet_authorship` and it implements like follow:

```rust
		// Check the memorized storage value.
		if let Some(id) = AccountId::<T>::get() {
			return id
		}

		let id = T::PalletId::get().into_account_truncating();
        AccountId::<T>::put(&id);
        id
```

la10736 (approved):
LGTM

PR #202, opened 2025-01-08
Risco0 WASM implementation

la10736:
Implement risc0 verifier in WASM using last release of risc0-verifier (v0.3.0) that use no-std. Also support multi verifier versions (1.0, 1.1, 1.2) for now. The native implementation also used the same crate instead of the original risc0 one.

The new version just support succinct and segment proof (not fake or groth16 ones) and use CBOR for serialization. 

1. The native implementation just recognize the fake receipts and return verify error
2. The new verifier use cbor for serialize proofs and it doesn't serialize the journal but just just expect the inner bytes instead
3. In the runtime script we should sanitize the environment by remove the `CARGO_FEATURE_*` envs injected by cargo that have some side effects on `cfg-aliases` crate used by `borsh-derive` 

p.s. we cannot use `bincode` for deserialization because doesn't support `no-std` compilation. CBOR is less efficient (since 20% bigger) but there are `no-std` implementations; `bincode` planed to implement the `no-std` support in the next `2.0.0` version, but even they already published some `dev` releases there isn't any ETA for the `2.0.0`. Anyway we can change it in the future by introduce a new variants for the `bincode` serialized proofs.


native/src/risc0.rs
```diff
@@ -17,26 +17,45 @@ use crate::VerifyError;
...
+        let journal = deserialize_pubs(pubs)?;
+
+        let ctx = risc0_verifier::VerifierContext::v1_0();
+        let proof = risc0_verifier::Proof::new(inner_receipt);
+        proof
+            .verify(&ctx, vk, journal.digest())
+            .map_err(|_| VerifyError::VerifyError)
+    }
+
+    /// Return if the proof is an Fake or Goth16 proof
```
DanieleDiBenedetto, line 50:
```suggestion
    /// Return if the proof is a Fake or a Groth16 proof
```

verifiers/risc0/src/lib.rs
```diff
@@ -89,6 +143,15 @@ impl<T: Config> Verifier for Risc0<T> {
     fn vk_bytes(_vk: &Self::Vk) -> hp_verifiers::Cow<[u8]> {
         panic!("Risc0 vk is already hashed and we cannot know its preimage: use vk_hash() instead")
     }
+
+    fn verifier_version_hash(proof: &Self::Proof) -> Option<H256> {
+        let h = match proof {
+            Proof::V1_0(_) => sp_io::hashing::sha2_256(b"risc0:v1.0"),
+            Proof::V1_1(_) => sp_io::hashing::sha2_256(b"risc0:v1.1"),
+            Proof::V1_2(_) => sp_io::hashing::sha2_256(b"risc0:v1.2"),
```
DanieleDiBenedetto, line 151: We can put directly the digests here and not recompute them every time

lgiussan (approved):
LGTM

PR #203, opened 2025-01-09
Add support for --password-filename to docker

drgora:
This PR updates the docker configuration for validators to allow using a password (optionally) for accessing the keystore.

This password is provided through a file, which is remove automatically once the node completes the startup.

Ultimately, this allows not keeping part of the secrets in cleartext on the validator filesystem.


DanieleDiBenedetto (approved):

docker/scripts/entrypoint.sh
```diff
@@ -232,5 +240,10 @@ if [[ (-n "${ZKV_CONF_BASE_PATH}") && (-n "${ZKV_CONF_CHAIN}") && (-f "${ZKV_NOD
   cp "${ZKV_NODE_KEY_FILE}" "${destination}/secret_ed25519"
 fi
 
+if [ -n "${ZKV_SECRET_PASSWORD_PATH}" ]; then
+    echo "Launching password file remover"
+    ./remove_password_file.sh ${ZKV_NODE} ${ZKV_SECRET_PASSWORD_PATH} &
```
lander86, line 245: Not a blocker : wouldn't it be better to use an absolute path ?

lander86 (approved):
LGTM, just one non blocking question

PR #206, opened 2025-01-16
Updated verifiers/ultraplonk to be use ultraplonk-no-std

DimitrisPapac:
This PR updates verifiers/ultraplonk to make use of the ultraplonk-no-std. All resource files were generated anew using Noir v.0.36.0 and were verified using noir-cli. Tests were also added to the benchmarking.rs module (inspired by their Risc0 counterparts).


verifiers/ultraplonk/src/lib.rs
```diff
@@ -65,12 +67,41 @@ impl<T: Config> Verifier for Ultraplonk<T> {
             hp_verifiers::VerifyError::InvalidInput
         );
 
-        log::trace!("Verifying (native)");
-        native::ultraplonk_verify::verify(*vk, proof, pubs).map_err(Into::into)
+        log::trace!("Verifying (no-std)");
+        ultraplonk_no_std::verify::<TestHooks>(vk, proof, pubs)
```
la10736, line 71: `TestHooks` is a bad name choice. WHat you can do is 

1. Call it with a more descriptive name like `CurveHooksImpl`.
2. Simply implement `CurveHooks` for `()` and use it `()` instead.

Also the name of `testhooks` module in the verifier crate is not a good choice. We'll handle it in the next iteration (next sprint) when we'll finalize the `ultraplonk_no_std` crate.

For now just change the local name: I left a comment on the `use` statement

verifiers/ultraplonk/src/lib.rs
```diff
@@ -17,12 +17,14 @@
...
 use sp_std::{marker::PhantomData, vec::Vec};
 
-pub use native::ULTRAPLONK_PROOF_SIZE as PROOF_SIZE;
-pub use native::ULTRAPLONK_PUBS_SIZE as PUBS_SIZE;
-pub use native::ULTRAPLONK_VK_SIZE as VK_SIZE;
+pub use ultraplonk_no_std::PROOF_SIZE;
+pub use ultraplonk_no_std::PUBS_SIZE;
+pub use ultraplonk_no_std::VK_SIZE;
+use ultraplonk_no_std::{key::VerificationKey, testhooks::TestHooks};
+
```
la10736, line 27:
```suggestion
use ultraplonk_no_std::{key::VerificationKey, testhooks::TestHooks as CurveHooksImpl};
```

verifiers/ultraplonk/src/lib.rs
```diff
@@ -65,12 +67,41 @@ impl<T: Config> Verifier for Ultraplonk<T> {
...
+                }
+                ultraplonk_no_std::errors::VerifyError::PublicInputError { message: _ } => {
+                    hp_verifiers::VerifyError::InvalidInput
+                }
+                ultraplonk_no_std::errors::VerifyError::KeyError => {
+                    hp_verifiers::VerifyError::InvalidVerificationKey
+                }
+                ultraplonk_no_std::errors::VerifyError::InvalidProofError => {
+                    hp_verifiers::VerifyError::VerifyError
+                }
```
la10736, line 88:
```suggestion
                    hp_verifiers::VerifyError::InvalidProofData
```

la10736: Also fix the test

la10736 (changes requested):
LGTM in general... just change the erro type in the case of invalid proof

verifiers/ultraplonk/src/lib.rs
DimitrisPapac: Sounds good to me.

DimitrisPapac:
> LGTM in general... just change the erro type in the case of invalid proof

Done.

la10736 (approved):
LGTM

lgiussan (approved):

PR #207, opened 2025-01-16
Update Ismp dependecy

la10736:
They broke the semantic versioning but maybe because the rid of the mmr dependecy. This PR is a starting draft... maybe @rdoria1 should look it in deep and understand better what is changed: maybe we can also rid off the mmr dependencies?


runtime/src/lib.rs
```diff
@@ -998,8 +997,8 @@ impl pallet_ismp::Config for Runtime {
     type Router = ModuleRouter;
     type Coprocessor = Coprocessor;
     type ConsensusClients = (ismp_grandpa::consensus::GrandpaConsensusClient<Runtime>,);
-    type Mmr = NoOpMmrTree<Runtime>;
     type WeightProvider = ();
+    type OffchainDB = ();
```
DanieleDiBenedetto, line 1001: Asked Hyperbridge guys and it seems it's fine to leave it like this for now

DanieleDiBenedetto (approved):

PR #210, opened 2025-01-21
Move docker creation in its own action and not workflow.

la10736:
Port on main and also fix a variable name. Even if the variable was missspelled 
it was not an issue because we use the default value


DanieleDiBenedetto (approved):

lander86 (approved):
LGTM

PR #213, opened 2025-01-22
Implemented the runtime migration that remove all ultraplonk vk and refunds the owners.

la10736:
No unit test implemented, but a check on runtime upgrade that the use are refunded.

I test the upgrade and it works, also the refound works. 

I DIDN'T TRY the try_runtime impl.

### DIFF
--- a/runtime/src/migrations.rs
+++ b/runtime/src/migrations.rs
@@ -18,6 +18,81 @@
 #[allow(unused_imports)]
 use super::*;
 
+mod migrate_change_vk_format {
+
+    use super::*;
+
+    pub struct RemoveUltraplonkVks;
+
+    #[cfg(feature = "try-runtime")]
+    type Balances = sp_std::collections::btree_map::BTreeMap<crate::AccountId, crate::Balance>;
+
+    impl RemoveUltraplonkVks {
+        #[cfg(feature = "try-runtime")]
+        fn get_all_tickets_balances() -> Balances {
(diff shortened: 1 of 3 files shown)


runtime/src/migrations.rs
```diff
@@ -18,6 +18,81 @@
...
+                "must upgrade linearly"
+            );
+
+            Ok(Self::get_all_tickets_balances().encode())
+        }
+
+        /// Migrates validators and nominators to bags list for the staking pallet.
+        fn on_runtime_upgrade() -> Weight {
+            use frame_support::traits::UncheckedOnRuntimeUpgrade;
+            if crate::System::last_runtime_upgrade_spec_version() >= 10_000 {
```
DanieleDiBenedetto, line 60: In order to speed things up I would integrate this migration in the runtime v0.10.0, so please update this number to 9000 instead

DanieleDiBenedetto (changes requested):
Must change tests in `verifiers/macros/src/lib.rs` to take into account the `Tickets `type that you've added in the generic verifier

runtime/src/migrations.rs
la10736: That's ok, it check if the version is >= than 10_000 in order TO SKIP IT (to avoid issue if we forget to remove it)

drgora (approved):
LGTM, also tested with `try-runtime` and all the tests pass.

DanieleDiBenedetto (approved):

PR #215, opened 2025-01-23
Fix rust docs

la10736:
1. We cannot document migration in pallet_verifier dua a bug in polkadot macro system (just skip it)
2. fix a missed link in pallet aggregate


drgora (approved):
LGTM

PR #217, opened 2025-01-23
Integrate try-runtime and zepter with the CI

drgora:
This PR updates the CI to:
- use `try-runtime` to execute an `on-runtime-upgrade` check of the updated runtime and its migrations against the state of the public testnet
- use zepter to check the correct propagation of crates' features and format the `Cargo.toml` files

Additionally, this PR removes the previous migration code used for updating the runtime from version 9000 to 10000 (as leaving it would result in failure), and integrates the changes suggested by zepter.


.github/workflows/CI-orchestrator.yml
```diff
@@ -50,7 +55,7 @@ jobs:
 
   set-overall-result:
     runs-on: ubuntu-latest
-    needs: [build-job, test-job, coverage-job, lint-format-job, zombienet-test-job, audit-job, check-act]
+    needs: [build-job, test-job, try-runtime-job, coverage-job, lint-format-job, zombienet-test-job, audit-job, check-act]
```
DanieleDiBenedetto, line 58: I was thinking to avoid increasing CI complexity (on top of everything we already have):

- We can move this job into the tag-orchestrator
- For the orchestrator we can make this job running only on main branch, and only after: merging OR schedule runs couple of times a week

.github/workflows/CI-lint-format.yml
```diff
@@ -54,6 +55,10 @@ jobs:
         shell: bash
         run: ${{ env.DOCKER_COMPOSE_CMD }} cargo fmt --check 2>&1 | tee formatting_output.txt
 
+      - name: Features
+        shell: bash
+        run: ${{ env.DOCKER_COMPOSE_CMD }} zepter run --config .cargo/zepter-ci.yaml | tee features_output.txt
```
DanieleDiBenedetto, line 60: Do we really need two separate config files for zepter ?

.github/workflows/CI-lint-format.yml
```diff
@@ -17,6 +17,7 @@ env:
   DOCKER_CARGO_HOME: /tmp/.cargo
   CMAKE_INSTALL: true
   LLD_INSTALL: true
+  CARGO_BINARIES_INSTALL: "zepter"
```
DanieleDiBenedetto, line 20: Considering zepter doesn't require any build, perhaps it could be worth it to move it to a separate job ?

DanieleDiBenedetto (changes requested):
Minor comments, but LGTM ! Thanks much !

.github/workflows/CI-try-runtime.yml
```diff
@@ -0,0 +1,57 @@
...
+  IMAGE_TAG: noble_rust-stable_latest
+  DOCKER_COMPOSE_CMD: "docker compose -f ${GITHUB_WORKSPACE}/ci/docker-compose.yml run --rm cargo-container"
+  DOCKER_BUILD_DIR: /build
+  DOCKER_CARGO_HOME: /tmp/.cargo
+  CMAKE_INSTALL: true
+  LLD_INSTALL: true
+
+jobs:
+  try-runtime:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
drgora: Given that this job does not perform the final linking for the node binaries, I thought we could stick with the smaller runners.

.github/workflows/CI-lint-format.yml
drgora: I'm afraid that's the best tradeoff. It looks like zepter is not really flexible in terms of cli, so that if you use a config file, you cannot extend the execution flags that from the command line. Given that we want different flags for the CI run and for the local run, and I want to keep the local usage as simple as executing `zepter` in the project root, I believe that having separate config files is the best solution.
An alternative would be to encode all the execution flags in the workflow yml, but I think that's less readable and maintainable.

DanieleDiBenedetto (approved):

la10736 (approved):
LGTM

PR #220, opened 2025-01-29
Halo2 verifier 

nulltea:
Uses: https://github.com/ChainSafe/halo2-verifier


verifiers/halo2/src/benchmarking.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+pub struct BenchData {
+    pub vk: Vec<u8>,
+    pub proof: Vec<u8>,
+    pub pubs: Vec<U256>,
+}
+
+pub fn valid_bench_data() -> BenchData {
+    let pubs_bytes = include_bytes!("resources/VALID_PUBS.bin").to_vec();
```
DanieleDiBenedetto, line 33: How complex is this proof used for benchmarking ? Keep in mind that these weights will be used to compute fees and to determine how many extrinsics validators can put in a block. So they mustn't be too small in case of complex proofs neither too high in case of simple proofs.
Take a look, for instance, at the work we did for benching the Risc0 and Ultraplonk verifier to get an idea

DanieleDiBenedetto (changes requested):
Thanks for the contribution ! Please integrate also the pallet and its weights in the runtime, and add tests of correct runtime inclusion and correct weights usage.
Take a look at other similar PRs to see how to do it.
Thanks !

DimitrisPapac (changes requested):
LGTM, overall. Just make sure to remove any commented-out code and make sure `clippy` is happy. :)

verifiers/halo2/src/benchmarking.rs
nulltea: Added benchmark routines for more complex proofs (k=8,16,21) similar to ultraplonk benchmarks

runtime/src/lib.rs
```diff
@@ -1072,6 +1079,7 @@ construct_runtime!(
         Ismp: pallet_ismp,
         IsmpGrandpa: ismp_grandpa,
         HyperbridgeAggregations: pallet_hyperbridge_aggregations,
+        SettlementFooPallet: pallet_halo2_verifier,
```
DanieleDiBenedetto, line 1082: typo

runtime/src/weights.rs
```diff
@@ -30,6 +30,7 @@ pub mod pallet_fflonk_verifier;
 pub mod pallet_grandpa;
 pub mod pallet_groth16_verifier;
 #[cfg(not(feature = "relay"))]
+pub mod pallet_halo2_verifier;
```
DanieleDiBenedetto, line 33: It should be present anyway, not gated behind this feature

verifiers/halo2/Cargo.toml
```diff
@@ -36,17 +36,13 @@ sp-io = { workspace = true, optional = true }
 pallet-aggregate = { workspace = true, optional = true, default-features = false, features = [
     "runtime-benchmarks",
 ] }
-halo2_verifier = { git = "https://github.com/ChainSafe/halo2-verifier.git", name = "halo2_verifier" }
-ff = { version = "0.13.0", default-features = false, features = ["alloc"] }
-serde = { version = "1.0.200", features = [
-    "derive",
-], optional = true, default-features = false }
+halo2_verifier = { git = "https://github.com/ChainSafe/halo2-verifier.git" }
```
DanieleDiBenedetto, line 39: please push a tag and reference it once ready

verifiers/halo2/src/benchmarking.rs
```diff
@@ -23,48 +21,94 @@ fn init<T: pallet_aggregate::Config>() -> (T::AccountId, u32) {
...
-        pubs.push(fr);
-    }
+    use super::*;
 
-    BenchData {
-        vk: include_bytes!("resources/VALID_VK.bin").to_vec(),
-        proof: include_bytes!("resources/VALID_PROOF.bin").to_vec(),
-        pubs,
+    #[benchmark]
+    fn submit_proof_8() {
```
DanieleDiBenedetto, line 34: Is the verifier complexity dependent solely on the number of public inputs ?

verifiers/halo2/src/verifier_should.rs
```diff
@@ -12,33 +12,16 @@ use super::*;
...
-    type PermutationMax = ConstU32<100>;
-    type SelectorMax = ConstU32<100>;
-    type LargestK = ConstU32<20>;
-    type QueriesMax = ConstU32<100>;
-    type ExpressionDegreeMax = ConstU32<100>;
-    type ExpressionVarsMax = ConstU32<100>;
-    type GatesMax = ConstU32<100>;
-    type LookupsMax = ConstU32<100>;
-    type ShufflesMax = ConstU32<100>;
+    type VkMaxBytes = ConstU32<750000000>;
```
DanieleDiBenedetto, line 15: Is this realistic or just for test ? What would be a realistic vk size ?

DanieleDiBenedetto (changes requested):

verifiers/halo2/src/benchmarking.rs
DanieleDiBenedetto: Also it would be nice to have an idea (we're going to run the benches) of the verification time as well as the proof and vk size

PR #226, opened 2025-02-11
For benchmarking generate genesis from wasm instead of chain spec

la10736:
Also add the possibility to choose what pallets use vis `PALLETS` env variable

### DIFF
--- a/scripts/bench_cfg.sh
+++ b/scripts/bench_cfg.sh
@@ -6,6 +6,7 @@ DEFAULT_BM_HEAP_PAGES=4096
 PROJECT_ROOT=${PROJECT_ROOT:-$(git rev-parse --show-toplevel)}
 WEIGTH_OUT_PATH=${WEIGTH_OUT_PATH:-""}
 ZKV_NODE_EXE=${ZKV_NODE_EXE:-"${PROJECT_ROOT}/target/production/zkv-node"}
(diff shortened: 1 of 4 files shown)


DanieleDiBenedetto (approved):

DanieleDiBenedetto:
@la10736 shouldn't we modify as well the script `reference_machine_run_all.sh` ?

la10736:
> @la10736 shouldn't we modify as well the script `reference_machine_run_all.sh` ?

For what? for the `PALLETS` configuration should already work like it is ... at least I think that I've tried it.... But I'll try it another time to be sure.

scripts/bench_cfg.sh
```diff
@@ -6,6 +6,7 @@ DEFAULT_BM_HEAP_PAGES=4096
 PROJECT_ROOT=${PROJECT_ROOT:-$(git rev-parse --show-toplevel)}
 WEIGTH_OUT_PATH=${WEIGTH_OUT_PATH:-""}
 ZKV_NODE_EXE=${ZKV_NODE_EXE:-"${PROJECT_ROOT}/target/production/zkv-node"}
+ZKV_RUNTIME=${ZKV_RUNTIME:-"${PROJECT_ROOT}/target/production/wbuild/zkv-runtime/zkv_runtime.wasm"}
```
drgora, line 9: I don't expect any concrete difference, but maybe we can be consistent and always use the compressed runtime blob?

drgora (approved):
LGTM

PR #227, opened 2025-02-11
Dimitris/remove fflonk, zksync, and old risc0 and ultraplonk

DimitrisPapac:
This PR is meant to remove support for the FFlonk and Zksync verifiers as required by tickets [ZKV-799](https://horizenlabs.atlassian.net/browse/ZKV-799?atlOrigin=eyJpIjoiN2QxYjk0MTkwMDdmNDFmYmExMzA1YjMzNzAzMGU1MmMiLCJwIjoiaiJ9) and [ZKV-800](https://horizenlabs.atlassian.net/browse/ZKV-800?atlOrigin=eyJpIjoiYmI0NjNkYjRmMGViNDdmZGEzMzg0OWI5Y2UyODhiMmYiLCJwIjoiaiJ9), respectively.
In addition, it removes the legacy UltraPlonk and Risc0 verifiers from `native`, as required by [ZKV-830](https://horizenlabs.atlassian.net/browse/ZKV-830?atlOrigin=eyJpIjoiYjdiMGQ1OWExYmZmNDZjZDlmZTMxNGJmMjJmOWNiNGQiLCJwIjoiaiJ9).


runtime/src/lib.rs
```diff
@@ -1064,9 +1044,7 @@ construct_runtime!(
         Offences: pallet_offences,
         Historical: pallet_session_historical::{Pallet},
         ImOnline: pallet_im_online,
-        SettlementFFlonkPallet: pallet_fflonk_verifier,
         Poe: pallet_poe,
-        SettlementZksyncPallet: pallet_zksync_verifier,
```
la10736, line 1069: I'm scared... I was sure that we put the indexes also in the solo-chain runtime.... But I was wrong. I'm not sure about what can be the clients side effect of this changes.

If we did this to be applied in the next regenesis maybe is better to take also the occasion of add also the indexes here (not in this PR ... indeed)

runtime/src/lib.rs
```diff
@@ -1173,8 +1151,8 @@ construct_runtime!(
         // Verifiers. Start indices at 160 to leave room and to the end (255). Don't add
         // any kind of other pallets after this value.
         CommonVerifiers: pallet_verifiers::common = 160,
-        SettlementFFlonkPallet: pallet_fflonk_verifier = 161,
-        SettlementZksyncPallet: pallet_zksync_verifier = 162,
+        // SettlementFFlonkPallet: pallet_fflonk_verifier = 161,
+        // SettlementZksyncPallet: pallet_zksync_verifier = 162,
```
la10736, line 1155: please remove the commented code. Moreover if this PR will be included in the regenesis release we can also push up the other verifiers indexes to close the hole. Ok, I thought about that: we can close the gap in any case because the relay chain is not released yet in any place

zombienet-tests/js_scripts/ultraplonk_data.js
la10736, line 1: Just curiosity: why did you change it?

la10736 (commented):
LGTM, just stay a little in stand-by in order to understand what we should do with the pallet indexes

la10736:
About the CI error: my guess is that because the runtime used to update the chain was build with the zksync native code inside. To fix it just rebuild a new runtime (use a huge spec version) and replace the old one used in the test.

BTW could be also useful also remove the barretenberger  native impl just to compile the runtime so this wasm could be used also for the future versions. Please contact me if you need any help to do it.

zombienet-tests/js_scripts/ultraplonk_data.js
DimitrisPapac: Wasn't sure if the data we had was valid (specifically, the length) for the current verifier. This is why I copy-pasted the data from the unit test in an effort to make the zombienet-test pass. Didn't work though.

la10736: If they pass before should pass also after your PR

runtime/src/lib.rs
DanieleDiBenedetto: We're going to get rid of everything related to solo chain and keep only relay, I think we can skip

zombienet-tests/js_scripts/0007-proof_with_vk.js
```diff
@@ -9,8 +9,8 @@ const ReturnCode = {
 };
 
 const { init_api, submitProof, registerVk, receivedEvents } = require('zkv-lib')
```
DanieleDiBenedetto, line 11: this `registerVk` function, coming from `zkv-lib/index.js` expects an event from settlementFflonkPallet. Please modify

DanieleDiBenedetto (changes requested):
Please do a CTRL+F in all project as there are other instances of fflonk and zksync. Moreover in the audit.toml file there are a couple of ignored findings that were related to a zksync dependency. Please remove them as well

zombienet-tests/js_scripts/0007-proof_with_vk.js
DimitrisPapac: Modified to use `settlementUltraplonkPallet` instead.

DanieleDiBenedetto (changes requested):
I'm afraid you also removed the new node-accelerated Risc0 verifier :D

runtime/Cargo.toml
```diff
@@ -93,8 +93,6 @@ aggregate-rpc-runtime-api = { workspace = true }
```
DanieleDiBenedetto, line 93: We can set the license to Apache-2 for the runtime, now that we have removed fflonk. Please also update the main readme accordingly.
PS. Also ultraplonk_verifier seems to have license GPL-3, which can be switched safely to Apache-2 now that we have removed barretenberg dependency

DanieleDiBenedetto (changes requested):

zombienet-tests/js_scripts/0007-proof_with_vk.js
DimitrisPapac: Done.

runtime/Cargo.toml
DimitrisPapac: Done.

README.md
```diff
@@ -156,18 +156,11 @@ The official documentation is available at [docs.zkverify.io](https://docs.zkver
 
 ## License
 
-zkVerify as a whole is released under the [GPL 3.0 license](LICENSE-GPL3). This is mostly due to the fact that the proof verification implemented by the `settlement-fflonk` pallet is based on GPL 3.0 software.
+zkVerify as a whole is released under the [Apache 2.0 license](LICENSE-APACHE2).
```
DanieleDiBenedetto, line 159: Please replace the whole License section with the following:

```suggestion
zkVerify is released under the following licenses:
- [GPL-3](LICENSE-GPL3) for `zkv-runtime` and `mainchain` crates.
- GPL-3 with CLASSPATH exception for `cumulus-relay-chain-inprocess-interface` and `cumulus-relay-chain-minimal-node` crates.
- [Apache 2.0](LICENSE-APACHE2) for all the other crates.
```

DanieleDiBenedetto (changes requested):

zombienet-tests/zkv-lib/index.js
```diff
@@ -166,7 +166,7 @@ async function waitForEvent(api, timeout, pallet, name) {
 
 exports.registerVk = async (pallet, signer, vk) => {
     return await submitExtrinsic(api, pallet.registerVk(vk), signer, BlockUntil.InBlock,
-        (event) => event.section == "settlementFFlonkPallet" && event.method == "VkRegistered"
+        (event) => event.section == "settlementUltraplonkPallet" && event.method == "VkRegistered"
```
la10736, line 169: Maybe that's just a _bug_ ... it's working just for ultraplonk... but it's still fine

la10736 (approved):
LGTM

PR #228, opened 2025-02-13
Refactored CI/CD and Docker Env

ysibirski:
This PR introduces major refactoring for CI/CD and Docker ENV

> ⚠️ 

- This PR introduces breaking changes for **compose-zkverify** project(no more _CONF_VAR_2. Entrypoint supports a list now)
- **Cache** per **MAIN** branch ONLY still due to current cache implementation on GH Actions level: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache
- Compose project needs to be updated once new image is built
- `--password-file` is going to be implemented in another PR
- Running CI locally tool `act` is being removed due to CI/CD workflow builds running in parallel vs sequential
- `staging/raw-chainspec.json` file is missing which is a requirement to run zkv-relay-docker-compose.yml project
- CI/CD build is being moved onto WarpBuild environment with unlimited cache and 50% off runners price vs GH Actions ones: https://www.warpbuild.com/products/ci-runners
- All build jobs can now be manually triggered as well


lander86 (approved):
LGTM . We'll have to fix --password-file logic in a later PR as just mentioned and another PR needs to follow this one to be sure to make compose repositories compatible.

docker/dockerfiles/zkv-node.Dockerfile
```diff
@@ -1,74 +1,75 @@
...
+      jq \
+    && useradd -m -U -s /bin/bash -d "/${RUN_USER}" "${RUN_USER}" \
+    && mkdir -p /data "/${RUN_USER}/.local/share" \
+    && chown -R "${RUN_USER}:${RUN_USER}" /data "/${RUN_USER}" \
+    && ln -s /data "/${RUN_USER}/.local/share" \
+    && apt-get -y clean \
+    && apt-get -y autoclean \
+    && apt-get -y autoremove \
+    && rm -rf /var/{lib/apt/lists/*,cache/apt/archives/*.deb} /tmp/*
+
```
la10736, line 67: Maybe move all these installation before copy the binary from the other layer generate a more stable cache that can be reused more times (on developer host of course)

docker/dockerfiles/zkv-relay.Dockerfile
```diff
@@ -1,76 +1,77 @@
...
+      curl \
+      jq \
+    && useradd -m -U -s /bin/bash -d "/${RUN_USER}" "${RUN_USER}" \
+    && mkdir -p /data "/${RUN_USER}/.local/share" \
+    && chown -R "${RUN_USER}:${RUN_USER}" /data "/${RUN_USER}" \
+    && ln -s /data "/${RUN_USER}/.local/share" \
+    && apt-get -y clean \
+    && apt-get -y autoclean \
+    && apt-get -y autoremove \
+    && rm -rf /var/{lib/apt/lists/*,cache/apt/archives/*.deb} /tmp/*
```
la10736, line 68: also here... maybe is better to move them at line 48

docker/scripts/entrypoint.sh
```diff
@@ -10,16 +10,14 @@
 # Environment variables in the form `ZKV_CONF_*` are translated to command line arguments based on these rules:
 #
 # 1. `ZKV_CONF_` prefix is removed
-# 2. if both a trailing underscore (`_`) and number are present, they are removed
```
la10736, line 13: Why are you removing this behavior? we introduced it for every cases where we need to map array flags: flags that can be used more than once like `--bootnodes`.

docker/scripts/entrypoint.sh
```diff
@@ -101,91 +107,95 @@ validate_and_download() {
...
-    if [ -n "${arg_value}" ]; then
-      conf_args+=("${arg_value}")
-    fi
-    echo "  ${var_name}=${var_value} -> ${arg_name} ${arg_value}"
+# Loop through all arguments to check for --dev
+for arg in "$@"; do
+  if [ "${arg}" == "--dev" ]; then
+    log_green "INFO: '--dev' flag detected! Running in development mode."
+    # You can place any logic for development mode here
+    DEV_MODE="true"
```
la10736, line 141: Nice

.github/workflows/CI-audit.yml
```diff
@@ -8,32 +8,61 @@ on:
...
-  IMAGE_TAG: noble_rust-stable_latest
-  DOCKER_COMPOSE_CMD: "docker compose -f ${GITHUB_WORKSPACE}/ci/docker-compose.yml run --rm cargo-container"
-  DOCKER_BUILD_DIR: /build
-  DOCKER_CARGO_HOME: /tmp/.cargo
-  CARGO_BINARIES_INSTALL: "cargo-audit"
-
 jobs:
   audit:
-    runs-on: ubuntu-latest
+    runs-on: warp-ubuntu-latest-x64-2x
```
la10736, line 13: That's a really fast job. Maybe can still run on github. What about to have 2 variables: one for fast job and other for heavy jobs?

.github/workflows/CI-audit.yml
```diff
@@ -8,32 +8,61 @@ on:
...
+  failure-notification:
+    runs-on: warp-ubuntu-latest-x64-2x
+    name: Notify if failed
+    needs: [audit]
+    if: ${{ failure() && !cancelled() && github.event_name == 'schedule' }}
+    steps:
+      - name: Send Slack notification
+        uses: slackapi/slack-github-action@v1.25.0
+        with:
+          payload: |
```
la10736, line 45: What about to do our own slack action that use the generic one but with just 2 arguments: `header` and `text`?

la10736: I' don't know if it is possible to have an array for `text` in order to cover the multiple sections scenario.

.github/workflows/CI-build.yml
```diff
@@ -1,82 +1,84 @@
...
+          use_cache: 'yes'
+          cache_key: 'build-mainchain'
+          cmake_install: 'yes'
+          lld_install: 'yes'
 
-      - name: Set docker env vars
-        run: |
-          echo "USER_ID=$(id -u)" >> $GITHUB_ENV
-          echo "GRP_ID=$(id -g)" >> $GITHUB_ENV
+      - name: Upload zkv-node binary artifact
```
DanieleDiBenedetto, line 26: The binary is pretty big, are we sure we want to upload it ?

.github/workflows/CI-build.yml
```diff
@@ -1,82 +1,84 @@
...
-      - name: Cargo build relay chain
-        run: ${{ env.DOCKER_COMPOSE_CMD }} cargo build -p zkv-relay --release
+          command: "cargo build -p zkv-relay -p paratest-node --release --features fast-runtime"
+          use_cache: 'yes'
+          cache_key: 'build-relay-paratest'
+          cmake_install: 'yes'
+          lld_install: 'yes'
 
-      - name: Upload relay chain runtime artifact
+      - name: Upload relay and paratest build binaries
```
DanieleDiBenedetto, line 50: Same comment as before, too big to upload them ?

.github/workflows/CI-build.yml
```diff
@@ -1,82 +1,84 @@
...
-            target/
-            deps/
-          key: build-cache
+          command: "cargo build -p zkv-runtime --release --features try-runtime"
+          use_cache: 'yes'
+          cache_key: 'build-try-runtime'
+          cmake_install: 'yes'
+          lld_install: 'yes'
+
+      - name: Upload try-runtime wasm artifact
```
DanieleDiBenedetto, line 78: No need for this

ci/docker.sh
la10736, line 1: It's not mater of your PR, but now also this name is misleading: it's just about publish image and extract wasm... not, like the name suggest, building image.

.github/workflows/CI-test.yml
```diff
@@ -4,83 +4,80 @@ run-name: "Workflow CI/CD Steps: build, unit and integration testing"
...
+      - name: Checkout code
+        uses: actions/checkout@v4
 
-      - name: Cargo unit tests
-        shell: bash
-        run: ${{ env.DOCKER_COMPOSE_CMD }} cargo test --lib --bins --all-features --no-fail-fast --release 2>&1 | tee unit_tests_output.txt
+      - name: Cargo unit test WITHOUT relay feature
+        uses: ./.github/actions/cargo-cmd
+        with:
+          command: "cargo test --lib --bins --all-features --no-fail-fast --release 2>&1 | tee unit_test_no_relay_output.txt"
```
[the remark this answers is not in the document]
DanieleDiBenedetto: I think you meant to pass `--all-features` ?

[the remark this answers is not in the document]
la10736: No, I meant exactly what I wrote, we're just testing `zkv-runtime` crate *without* the `relay` feature and with `runtime-feature`. When we tested all crates with all features the runtime was tested with also the `relay` features activated, but when we use the `zkv-runtime` in solo chain the feature is disabled and the runtime is really different.

.github/actions/cargo-cmd/action.yml
```diff
@@ -0,0 +1,119 @@
...
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+          deps/
+        key: ${{ inputs.cache_key }}-${{ github.event.repository.default_branch }}
+        delete-cache: true
```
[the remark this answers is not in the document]
ysibirski: Yes, unfortunately GH Actions cache implementation enforce CACHE_KEY to be immutable, therefore if we want to overwrite the existing cache key coming from the same branch we have to DELETE it first and WRITE again.

.github/workflows/CI-build-test-publish.yml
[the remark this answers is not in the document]
ysibirski: Agreed, I was thinking about changing this as well. Will take care of it

.github/actions/cargo-cmd/action.yml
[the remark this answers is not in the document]
ysibirski: Addressed

.gitignore
```diff
@@ -4,6 +4,7 @@ target/
 *.iml
 .secrets
 .vscode
+actionlint.yaml
```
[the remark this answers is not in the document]
ysibirski: I think it is a very good idea for all of US working on the pipeline to use linter. I have discovered it and it looks/works very good. Even, if we include actionlint.yaml under gitignore it is not enforcing its usage, but removes the headache of someone who will actually use it.

I am indifferent on this one. @la10736 let me know what do you prefer. I can always add this file and delete it all the time I am working on the pipeline, but thought keep it here will not hurt :shrug:

docker/dockerfiles/zkv-node.Dockerfile
```diff
@@ -1,74 +1,75 @@
...
 
 ARG PROFILE="release"
 ARG FEATURES=""
 
 WORKDIR /usr/src/node
 COPY . .
 
-RUN echo "SUBSTRATE_CLI_GIT_COMMIT_HASH=`git rev-parse --short=11 HEAD`" >> .docker.env
-RUN . ./.docker.env \
-  && cargo build -p mainchain --profile ${PROFILE} --features "${FEATURES}"
```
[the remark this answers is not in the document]
ysibirski: I think I was not sure how do we use this `SUBSTRATE_CLI_GIT_COMMIT_HASH` var?  Is it actually being used for `cargo build` command? Tell me more about this please.

.github/workflows/CI-build.yml
la10736: We need them for zombienet. Don't it?

docker/dockerfiles/zkv-node.Dockerfile
ysibirski: Good point

.github/workflows/CI-build.yml
la10736: I guess that here we already have a lot of stuff cached ... Rebuilt it on `CI-try-runtime` means recompile them. Anyway it makes the sequential... but the just try step without compile is really fast (2 minutes) so maybe this approach is a little better

.github/workflows/CI-audit.yml
ysibirski: We are running this on SMALLER instance anyway and paying 75% less then if we would run on GH Actions default runner.

.github/workflows/CI-audit.yml
ysibirski: That is a good idea, but to be honest with you IF we go this path, we should deploy it as an action that can be shared by ALL our repos, therefore thinking on creating centralized GH Actions repo and then using those in our workflows since we are using this `SLACK-github-action` under each of the pipeline we wrote.

I would leave this improvement for future task since we probably want to think on how this should be implemented :shrug:

.github/workflows/CI-build.yml
DanieleDiBenedetto: Of course, but why uploading them ? Can't we save them locally somewhere ?

.github/workflows/CI-build.yml
DanieleDiBenedetto: Of course, but why uploading them ? Can't we save them locally somewhere ?

.github/workflows/CI-build.yml
ysibirski: That is the only way to GET the artifact using a DIFF workflow/job. Artifacts space is free but has some limits like single artifact file can not exceed 2 GB and etc. In our case we use default compression level = 6 and its total size is 21MB

.github/workflows/CI-build.yml
ysibirski: Adjusting

ci/docker.sh
ysibirski: Renaming to publish-docker-image instead

.gitignore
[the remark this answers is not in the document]
ysibirski: Adding

docker/scripts/entrypoint.sh
ysibirski: This is not being removed, rather being improved. With new implementation you no longer need to pass ONE value per VARIABLE if it supports list. Now you can do the following:

OLD
```
ZKV_CONF_LISTEN_ADDR="/ip4/0.0.0.0/tcp/${NODE_NET_P2P_PORT}"
ZKV_CONF_LISTEN_ADDR_2="/ip4/0.0.0.0/tcp/${NODE_NET_P2P_PORT_WS}/ws"
```

NEW
```
ZKV_CONF_LISTEN_ADDR="/ip4/0.0.0.0/tcp/${NODE_NET_P2P_PORT},/ip4/0.0.0.0/tcp/${NODE_NET_P2P_PORT_WS}/ws"
```

.github/workflows/CI-build.yml
ysibirski: Same answer as before. It is ok to upload them as artifacts, stored for 1 day with default compression level 6

image

ysibirski:
@la10736 and @DanieleDiBenedetto please read all my replies and take a look at my latest commit where I addressed all the comments I had not further questions or explanation about

docker/dockerfiles/zkv-node.Dockerfile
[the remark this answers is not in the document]
la10736: Ok, you're right!!! When I added this code to solve the issue I did 2 changes

- this line
- removed from docker ignore file the rows about ignore `.git` and `.gitignore` 

Just the second was enough to solve the issue!!!

The code used in `build.rs` script via`generate_cargo_keys()` https://docs.rs/substrate-build-script-utils/latest/src/substrate_build_script_utils/version.rs.html#21-50 do exactly this one if the the variable isn't defined.

Thanks again

ci/setup_env.sh
```diff
@@ -26,21 +26,22 @@ fi
...
 import_gpg_keys() {
   # shellcheck disable=SC2207
-  declare -r my_arr=( $(echo "${@}" | tr " " "\n") )
+  declare -r keys=( $(echo "${@}" | tr " " "\n") )
 
-  if [ "${#my_arr[@]}" -eq 0 ]; then
-    log_warn "WARNING: there are ZERO gpg keys to import. Please check if MAINTAINERS_KEYS variable(s) is(are) set correctly. The build is not going to be released ..."
-    export IS_A_RELEASE="false"
+  if [ "${#keys[@]}" -eq 0 ]; then
+    log_warn "WARNING: there are ZERO gpg keys to import. Please check if 'MAINTAINERS_KEYS' variable is set correctly. The build is not going to be released ..."
```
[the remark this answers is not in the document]
la10736: Ok,we'll do it in a next PR

docker/scripts/entrypoint.sh
la10736: Ok, I got it. We chose the other solution because it also possible to use it when the arguments can contain also `,`. But it's ok for now.... if we'll meet some issue we'll change it later.

DanieleDiBenedetto (approved):

la10736 (approved):
LGTM

Could you run it on the other repo before merge it?

docker/README.md
```diff
@@ -17,7 +17,7 @@ Start a _**solo**_ chain with 2 validators (Alice and Bob) and a simple client n
 > docker compose -f docker/dockerfiles/zkv-docker-compose.yaml up
 ```
 
-Start a _**relay**_ chain with a test _**parachain**_ with:
+**TO BE SUPPORTED:** Start a _**relay**_ chain with a test _**parachain**_ with:
```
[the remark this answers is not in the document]
ysibirski: Yes, it requires spec file under ../../staging dir and there is nothing there, but perhaps it will be worth mentioning that file needs to be placed there?

PR #233, opened 2025-02-17
Update consensus params

drgora:
This PR updates many consensus parameters. See individual commit titles for a quick overview.
Tests are updated accordingly.

### DIFF
--- a/relay-node/service/src/chain_spec.rs
+++ b/relay-node/service/src/chain_spec.rs
@@ -250,9 +250,9 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
         WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?,
         Default::default(),
     )
-    .with_name("ZKV Testnet")
+    .with_name("zkVerify Testnet")
     .with_id("zkv_testnet")
-    .with_protocol_id("tacme")
+    .with_protocol_id("tzkv")
     .with_chain_type(ChainType::Live)
     .with_boot_nodes(vec![
         format!("/dns/{BOOTNODE_1_DNS}/tcp/30333/p2p/{BOOTNODE_1_PEER_ID}")
--- a/relay-node/service/src/chain_spec.rs
+++ b/relay-node/service/src/chain_spec.rs
@@ -210,7 +210,14 @@ pub fn development_config() -> Result<ChainSpec, String> {
                 ),
             ])
             .collect::<Vec<_>>(),
-        true,
+        // min validator count
+        1,
+        // max validator count
+        None,
+        // min validator bond
+        0,
+        // min nominator bond
+        0,
     ))
     .build())
 }
@@ -239,7 +246,14 @@ pub fn local_config() -> Result<ChainSpec, String> {
             .into_iter()
             .map(|seed| (get_account_id_from_seed::<sr25519::Public>(seed), ENDOWMENT))
             .collect::<Vec<_>>(),
-        true,
+        // min validator count
+        1,
+        // max validator count
+        None,
+        // min validator bond
+        0,
+        // min nominator bond
+        0,
     ))
     .build())
 }
@@ -254,6 +268,7 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
     .with_id("zkv_testnet")
     .with_protocol_id("tzkv")
     .with_chain_type(ChainType::Live)
+    /*
     .with_boot_nodes(vec![
         format!("/dns/{BOOTNODE_1_DNS}/tcp/30333/p2p/{BOOTNODE_1_PEER_ID}")
             .parse()
@@ -274,6 +289,7 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
             .parse()
             .expect("MultiaddrWithPeerId"),
     ])
+    */
     .with_telemetry_endpoints(
         TelemetryEndpoints::new(vec![(
             STAGING_TELEMETRY_URL.to_string(),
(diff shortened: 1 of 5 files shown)


runtime/src/tests/pallets_interact.rs
```diff
@@ -21,55 +21,18 @@ use frame_support::{
...
 use sp_runtime::{
     traits::{Hash, Header as _},
-    AccountId32, Digest, DigestItem,
+    AccountId32, DigestItem,
 };
-use testsfixtures::get_from_seed;
 
 use super::*;
 
-// Any random values for these constants should do
```
DanieleDiBenedetto, line 35: Why you removed this test ?

runtime/src/tests/payout.rs
```diff
@@ -15,8 +15,10 @@
 
 //! This module tests the correct computation of rewards for validators.
 
-use crate::{payout::*, Balance, Perbill, Runtime, ACME, MILLICENTS};
+use super::*;
```
DanieleDiBenedetto, line 18: All these tests are neat, but I was wondering if there is a way to expose some adjustable constants, so that we don't have to modify manually all the tests in future, when (and if) we are going to change inflation and reward distribution rules

DanieleDiBenedetto (approved):
Awesome job ! Left a couple non-blocking comments.
I was wondering if it could be worth tho also put in this PR the genesis config presets for all the affected pallets (according to our document)

runtime/src/lib.rs
```diff
@@ -614,7 +633,7 @@ impl pallet_aggregate::Config for Runtime {
     type RuntimeHoldReason = RuntimeHoldReason;
     type AggregationSize = AggregateMaxSize;
     type MaxPendingPublishQueueSize = AggregateQueueSize;
-    type ManagerOrigin = EnsureRoot<AccountId>;
+    type ManagerOrigin = EitherOfDiverse<EnsureRoot<AccountId>, Treasurer>;
```
DanieleDiBenedetto, line 636: As of now, perhaps, it doesn't make much sense to have `Treasurer` here...
@la10736 what do you think ?

la10736: I don't think `Treasurer` is correct here.: the manager here is a just a technical maintainer that eventually can solve issues. Also in a future where we'll remove `sudo` and use governace for runtime upgrade I guess that here we'll have a multiginature account instead.

runtime/src/tests.rs
```diff
@@ -16,9 +16,8 @@
 #![cfg(test)]
 
 use super::*;
-
 /// Reexport `test` runner
```
la10736, line 19: Is the comment still correct?

la10736 (approved):
LGTM... just a minor note

runtime/src/lib.rs
drgora: Alright, reverted the change

runtime/src/payout.rs
```diff
@@ -102,10 +113,14 @@ impl pallet_staking::EraPayout<Balance> for ZKVPayout {
         let s = StakingTarget::get().saturating_reciprocal_mul(staking_current * SCALE as u128);
 
         // exp_arg = k * (1 - s)
-        let exp_arg = (SCALE - s as f64) * K::get() / SCALE;
-
-        // inflation_arg = C * e^(exp_arg)
-        let inflation_arg = C::get() * exp(exp_arg);
+        let inflation_arg = if C::get() != 0f64 {
```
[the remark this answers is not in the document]
drgora: Yes I understand your concern, but this is used only for this very specific condition, and we will probably remove it once the tokenomics will be settled.

runtime/src/lib.rs
```diff
@@ -720,14 +739,17 @@ parameter_types! {
 pub const MAX_TARGETS: u32 = 1_000;
 // Maximum number of voters. This also includes targets, which implicitly vote for themselves.
 pub const MAX_VOTERS: u32 = 5_000;
+// The maximum number of number of active validators that we want to handle.
+// This *must always be greater or equal* to staking.validatorCount storage value.
+pub const MAX_ACTIVE_VALIDATORS: u32 = 20;
```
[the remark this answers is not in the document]
drgora: Yes, correct.

relay-node/service/src/chain_spec.rs
```diff
@@ -250,10 +264,11 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
         WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?,
         Default::default(),
     )
-    .with_name("ZKV Testnet")
+    .with_name("zkVerify Testnet")
     .with_id("zkv_testnet")
-    .with_protocol_id("tacme")
+    .with_protocol_id("tzkv")
     .with_chain_type(ChainType::Live)
+    /*
```
DanieleDiBenedetto, line 271: why you commented this ?

drgora: To be able to generate a test chain spec to check the changes in this PR (waiting for the network keys of the regenesis bootnodes); but ok, I can remove the comment in the meantime

relay-node/service/src/chain_spec.rs
```diff
@@ -294,7 +310,14 @@ pub fn testnet_config() -> Result<ChainSpec, String> {
         vec![
             // TBD
```
DanieleDiBenedetto, line 311: I think here we can keep same addresses and allocations as in the current testnet

DanieleDiBenedetto (approved):

PR #235, opened 2025-02-24
Dimitris/version in statement hash

DimitrisPapac:
This PR addresses the [ZKV-832](https://horizenlabs.atlassian.net/browse/ZKV-832?atlOrigin=eyJpIjoiZmQzZmYyOTdjOGNkNGJlOGE4ZWE3MDA1YzkwMWJjNmMiLCJwIjoiaiJ9) ticket by hashing an empty slice when no version is provided. A corresponding unit test is also added.


DanieleDiBenedetto (changes requested):
Perhaps we could something more modular and future proof, as I was thinking what it would look like if the verifiers that currently have no version will start to have one. Two alternatives here: 
- Supply a default implementation to the function `verifier_version_hash` in `primitives/hp_verifiers/src/lib.rs` returning the `no_version_hash`, and overwrite it once a verifier will start to be versioned
- Define a version for all the verifiers (we can set whatever we want) and produce the hash accordingly.
In any case, the `verifier_version_hash` function shouldn't return an `Option` anymore.

@la10736 @lgiussan what do you think ?

pallets/verifiers/src/lib.rs
```diff
@@ -193,6 +198,8 @@ pub mod pallet {
         data_to_hash.extend_from_slice(vk_hash.as_bytes());
         if let Some(h) = version_hash {
             data_to_hash.extend_from_slice(h.as_bytes());
+        } else {
+            data_to_hash.extend_from_slice(&NO_VERSION_HASH);
         }
```
lgiussan, line 203: I agree with DDB's comment: let's change the signature of `hp_verifiers::Verifier::verifier_version_hash` to always return a `H256` (and not an `Option<H256>`) and make the default implementation return `NO_VERSION_HASH`. Of course `NO_VERSION_HASH` definition should be moved there too, and I would change it to be the hash of the empty string.

pallets/verifiers/src/lib.rs
```diff
@@ -493,14 +501,18 @@ pub mod pallet {
             0,
             42,
             VkOrHash::from_vk(REGISTERED_VK),
-            VALID_HASH_REGISTERED_VK
+            H256(hex!(
+                "a3f38d8db1ff6c75d6977b5a5b23a375df6f1defc9bd13742d50e5547ab02b4f"
+            ))
```
lgiussan, line 506: Maybe you could define a constant, instead of copy-pasting this value multiple times.

pallets/verifiers/src/lib.rs
DimitrisPapac: Done.

pallets/verifiers/src/lib.rs
DimitrisPapac: Done.

lgiussan (approved):
LGTM!

la10736:
> Perhaps we could something more modular and future proof, as I was thinking what it would look like if the verifiers that currently have no version will start to have one. Two alternatives here:
> 
> * Supply a default implementation to the function `verifier_version_hash` in `primitives/hp_verifiers/src/lib.rs` returning the `no_version_hash`, and overwrite it once a verifier will start to be versioned
> * Define a version for all the verifiers (we can set whatever we want) and produce the hash accordingly.
> 
> In any case, the `verifier_version_hash` function shouldn't return an `Option` anymore.
> 
> @la10736 @lgiussan what do you think ?

My vote is for the first option

PR #236, opened 2025-02-27
Risc0 verifier dynamic weights

la10736:


verifiers/risc0/src/lib.rs
```diff
@@ -32,25 +34,41 @@ pub use fake_extend_benchmarking as extend_benchmarking;
...
-        Self::MaxProofSize::get()
+        Self::MaxNSegment::get() * Self::Segment20MaxSize::get()
     }
 
     fn max_pubs_size() -> u32 {
         Self::MaxPubsSize::get()
     }
+
+    fn max_verify_proof_weight() -> Weight {
+        Self::WeightInfo::verify_proof_segment_poseidon2_16() * Self::MaxNSegment::get() as u64
```
lgiussan, line 66: Shouldn't we use `verify_proof_segment_poseidon2_20()`?
```suggestion
        Self::WeightInfo::verify_proof_segment_poseidon2_20() * Self::MaxNSegment::get() as u64
```

primitives/hp-verifiers/src/lib.rs
```diff
@@ -64,7 +64,7 @@ pub trait Verifier: 'static {
         vk: &Self::Vk,
         proof: &Self::Proof,
         pubs: &Self::Pubs,
-    ) -> Result<(), VerifyError>;
+    ) -> Result<Option<Weight>, VerifyError>;
```
DanieleDiBenedetto, line 67: Update doc please

primitives/hp-verifiers/src/lib.rs
```diff
@@ -144,7 +144,7 @@ impl Verifier for () {
         _vk: &Self::Vk,
         _proof: &Self::Proof,
         _pubs: &Self::Pubs,
-    ) -> Result<(), VerifyError> {
+    ) -> Result<Option<Weight>, VerifyError> {
```
DanieleDiBenedetto, line 147: Update doc please

pallets/verifiers/src/lib.rs
```diff
@@ -343,18 +345,22 @@ pub mod pallet {
...
+            Self::deposit_event(Event::ProofVerified { statement });
+            T::OnProofVerified::on_proof_verified(account, domain_id, statement);
+            Ok(verify_proof_weight
+                .map(|new_weight| {
+                    submit_proof_weight::<T, I>(
+                        &vk_or_hash,
+                        &proof,
+                        &pubs,
+                        &domain_id,
+                        Some(new_weight),
```
DanieleDiBenedetto, line 360: Minor: a little bit difficult to follow the logic here but I understand there's not much we can do differently

DanieleDiBenedetto (changes requested):
Minor stuff, LGTM for the rest !

verifiers/risc0/src/lib.rs
la10736: Yes, you're right... but the real question here is ... why there isn't any tests that catch it?

primitives/hp-verifiers/src/lib.rs
la10736: I guess the `trait`'s one is enough

DanieleDiBenedetto (approved):

lgiussan (approved):

PR #241, opened 2025-03-04
Docker entrypoint adjustment for secrets

ysibirski:
This PR introducing the following:

- GOSU for zkVerify image
- NODE_KEY and SECRET_PHRASE are coming from ENV vars now vs file on the host filesystem (will be imported from Vault for OUR deployments)
- NODE_KEY and SECRET_PHRASE will be stored on TMPFS path `/tmp` location (coming from compose project)
- `password-file` option is removed since secrets will be coming from Vault under compose project for our deployments. This password file hard to manage since it has to be deleted after application start and needs to be present on application restart
- entrypoint.sh takes care of secret files creation if ENV vars are provided
- Download spec file function was improved by adding retry logic in case download command fails

@la10736 and/or @drgora help me with `.env.local` files coming from testing compose project. I am not sure where `node_key.dat` were coming from. In my PR the value was deleted and var name renamed.


lander86 (approved):
It looks good to me, everything is as already discussed 💯 . Let's wait for @drgora and @la10736 review. Changes in the compose repositories will follow.

docker/resources/envs/.env.local
```diff
@@ -1,11 +1,11 @@
 # RUST_LOG=debug
 
-ZKV_NODE_KEY_FILE="/data/config/node_key.dat"
```
drgora, line 3: To answer your question, the node key is randomly generated, just like any other secret.
Maybe we should also add `ZKV_SECRET_PHRASE` in these templates (possibly with some predefined value like `"YOUR_SECRET_PHRASE"`), to make it clear where it should be provided?

ysibirski: I will add this to each `.env.local` file where ZKV_NODE_KEY is present with empty "" value to make sure if someone forgets to change COMMENT value entrypoint script will not fail since it is processing these vars now

ysibirski:
> Thank you for the PR! I am most likely missing something here, but there are some points I don't understand.
> 
> In the current status, it looks to me that the node secrets will always be stored in cleartext in the container (regardless of the files in `/tmp`, you also have it in the keystore of the node). If this is correct, I do not see the point of using HC Vault for passing such secrets to the container.

The reason and idea to inject secrets from HC vault is that those secrets WILL never be exposed on the hosts FILESYSTEM. Therefore, you can only get to those secrets IF you have SUDO rights and/or part of `docker` group. We are saving the secrets inside docker container on `/tmp/FILE` which is TMPFS volume = RAM vs DISK and it is OWNED by RUNUSER with 0400 permissions = read-only for RUN_USER

@drgora let me know if this all make sense and/or if you have more questions. I can provide more explanations for this

docker/scripts/entrypoint.sh
```diff
@@ -24,6 +24,8 @@
 
 set -eEuo pipefail
 
+RUN_USER="${RUN_USER:-}"
```
cronicc, line 27: Maybe fail early here if unset or empty. Otherwise we'd only fail later with `unbound variable`.

docker/scripts/entrypoint.sh
```diff
@@ -148,10 +167,27 @@ if [ "${DEV_MODE:-false}" != "true" ]; then
   validate_and_download "ZKV_CONF_CHAIN" "ZKV_SPEC_FILE_URL"
 fi
 
-ZKV_SECRET_PHRASE_PATH=${ZKV_SECRET_PHRASE_PATH:-"/data/config/secret_phrase.dat"}
-ZKV_NODE_KEY_FILE=${ZKV_NODE_KEY_FILE:-"/data/config/node_key.dat"}
-echo "  ZKV_SECRET_PHRASE_PATH=${ZKV_SECRET_PHRASE_PATH}"
-echo -e "  ZKV_NODE_KEY_FILE=${ZKV_NODE_KEY_FILE}\n"
+# Creating node key file if node key is provided
+if [ -n "${ZKV_NODE_KEY}" ]; then
+  ZKV_NODE_KEY_FILE="/tmp/node_key.dat"
```
cronicc, line 172: Maybe we'd like to have the option of this path being user controlled. You could get the value from the env var if set and alternatively set a default path on L151.

/tmp is also not guaranteed to be in memory, by default with Docker it is disk backed. See:
```
docker run --rm ubuntu:jammy findmnt -n -o SOURCE --target /tmp/
/dev/sdf1[/subvol/var/lib/docker/btrfs/subvolumes/a77d7edc4c8e51cbadb906fe0c4b0e88208c9e13027889bee81b6348af56da87]
```

`/dev/shm` would be the safer alternative if we never want to write this file to disk.

docker/scripts/entrypoint.sh
```diff
@@ -148,10 +167,27 @@ if [ "${DEV_MODE:-false}" != "true" ]; then
...
+  printf "%s" "${ZKV_NODE_KEY}" > "${ZKV_NODE_KEY_FILE}"
+  chmod 0400 "${ZKV_NODE_KEY_FILE}" && chown "${RUN_USER}" "${ZKV_NODE_KEY_FILE}"
+  echo -e "  ZKV_NODE_KEY_FILE=${ZKV_NODE_KEY_FILE}\n"
+
+  unset ZKV_NODE_KEY
+fi
+
+# Creating secret phrase file if secret phrase is provided
+if [ -n "${ZKV_SECRET_PHRASE}" ]; then
+  ZKV_SECRET_PHRASE_FILE="/tmp/secret_phrase.dat"
```
cronicc, line 183: See above.

docker/scripts/entrypoint.sh
ysibirski: Sounds good.

docker/scripts/entrypoint.sh
ysibirski: I was thinking of more enforcing this location to be under `tmpf` volume which will be manadatory added to the compose project

```    
tmpfs:
      - /tmp
 ```
Therefore, not giving USER's an option to save it somewhere else rather than RAM DISK.

@cronicc let me know what do you think

File: docker/scripts/entrypoint.sh (hunk 4 above)
Comment (ysibirski) [reply docker/scripts/entrypoint.sh:183]: Discussed in person. Switching to `/dev/shm/` location for both secrets without option to change the location to enforce RAMDISK

**Comment by ysibirski** [comment]
@cronicc @drgora @lander86 I have addressed all the comments and here is the list of improvements

- Node key, secret phrease and keystore are now being saved on `/dev/shm` = RAMDISK enforced location even if someone tries to run it without compose project
- Node key no longer needed to be saved under `base-path/..../network/*` location since it is being stored on RAMDISK
- keystore is only being created if SECRET_PHRASE is passed. Node key is not using keystore

Please let me know what do you think.

File: docker/scripts/entrypoint.sh [hunk 5]
```diff
@@ -183,22 +200,36 @@ while IFS='=' read -r -d '' var_name var_value; do
...
+  printf "%s" "${ZKV_NODE_KEY}" > "${ZKV_NODE_KEY_FILE}" || fn_die "ERROR: was not able to create '${ZKV_NODE_KEY_FILE}' file.  Exiting ..."
+  chmod 0400 "${ZKV_NODE_KEY_FILE}" && chown "${RUN_USER}" "${ZKV_NODE_KEY_FILE}"
+  echo -e "  ZKV_NODE_KEY_FILE=${ZKV_NODE_KEY_FILE}\n"
+
+  conf_args+=(--node-key-file "${ZKV_NODE_KEY_FILE}")
+  unset ZKV_NODE_KEY
+fi
+
+# Creating secret phrase file and keystore location on RAMDISK if secret phrase is provided
+if [ -n "${ZKV_SECRET_PHRASE}" ]; then
```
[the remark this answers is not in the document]
ysibirski: Sounds good @drgora., I will work on the implementation of this

ysibirski:
@drgora @cronicc the latest commit add the following:

- Saving secret_phrase and node_key on docker volume by default, but can be overwritten via env var(will be using RAMDISK for our deployments)
- Secret_phrase is a REQUIREMENT if running a validator node and entrypoint will exit preventing applicatio from running
- If node_key is not provided enrypoint will generate ONE and save it on docker volume. As long as this file exist on docker volume entrypoint will NOT generate new node key and preserve it.
- Keystore is saved under specified path as well vs default dir

I am still performing more tests, but please review my latest commit at your earliest convenience
Thank you in advance.

docker/scripts/entrypoint.sh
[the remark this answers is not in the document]
ysibirski: Added all the checks and generating RANDOM node key inside the entrypoint if not provided

drgora (approved):
Looks good to me, thanks!

lander86 (approved):
All the improvements are looking good to me. Thanks @ysibirski

cronicc (approved):
LGTM

ysibirski:
- The logic was improved by injecting session keys ONLY if `SECRET_PHRASE` var is provided,
- Injection of the keys was moved under for loop

cronicc (approved):

PR #244, opened 2025-03-11
Enforce locked Cargo.lock in CI

lgiussan:
At the moment, CI doesn't catch possibly broken `Cargo.lock` lockfiles, because the default behavior of `cargo check`, `cargo build`, etc. is to silently fix problems inside the `Cargo.lock`. This issue can arise especially after rebases and merges.

The problem could be prevented by simply passing the `--locked` flag to the various `cargo check`s, and `cargo build`s used throughout the CI. However I decided to add a dedicated CI step that only targets this particular check, so that a developer can have a clearer feedback on why CI fails.

### DIFF
--- a/.github/workflows/CI-check-lockfile.yml
+++ b/.github/workflows/CI-check-lockfile.yml
@@ -1,0 +1,21 @@
+name: Check Cargo.lock
+
+run-name: "Workflow CI/CD Step: check Cargo.lock"
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  check-cargo-lock:
+    runs-on: warp-ubuntu-latest-x64-2x
+    name: Check lockfile
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Check lockfile
+        uses: ./.github/actions/cmd-in-docker
+        with:
+          command: "cargo update --workspace --locked"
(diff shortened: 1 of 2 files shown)


ysibirski (approved):
LGTM, Just one question, does this new pipeline workflow need to run in parallel with the other workflows or it may be a requirement for all the others to run after?

lgiussan:
> LGTM, Just one question, does this new pipeline workflow need to run in parallel with the other workflows or it may be a requirement for all the others to run after?

Logically, it could make sense to put this new workflow as a requirement for all the others. However, in the happy path, this could introduce more than 1 min of delay to the overall running time.

lander86:
The PR looks good to me. The 1 min extra time will absolutely be not a problem as pointed out by @ysibirski already. Let's only understand if we have to fail the whole pipeline should this step fail @lgiussan .

lander86 (approved):
Please answer to the last comments @lgiussan

lgiussan:
@ysibirski @lander86 Thanks for raising the issue. In commit https://github.com/zkVerify/zkVerify/pull/244/commits/2e8e8509b3e5558d876264d3a387cba6e4698e01 I made the lockfile check a prerequisite for all the other workflows.

DanieleDiBenedetto (approved):

PR #245, opened 2025-03-11
Relay Chain

drgora:
This PR removes the solo chain node, leaving only the relay chain implementation.

Other notable changes:
- removed proof_of_existence pallet and RPC
- upgraded Polkadot dependencies to version stable-2409-5
- updated consensus parameters
- removed fflonk, zksync, and old risc0 and ultraplonk 
- removed unneeded governance tracks


.github/workflows/CI-try-runtime.yml
```diff
@@ -9,14 +9,14 @@ on:
         description: 'Whether to execute try-runtime step or not'
         required: false
         type: string
-        default: 'yes'
+        default: 'no'
```
ysibirski, line 12: If default is `no` now you may want to set it to `yes` under orchestrator workflow?

```
  try-runtime-job:
    name: Try runtime job
    uses: ./.github/workflows/CI-try-runtime.yml
```

docker/README.md
```diff
@@ -93,28 +79,14 @@ All Dockerfiles are located in `docker/dockerfiles` folder.
 
 * `zkv-builder.Dockerfile`: create an image with all dependencies needed to compile the node and is used by `my_cargo` script
 * `binary_injected.Dockerfile`: Is mainly used by the scripts and inject one or more binaries in a standard ubuntu image
-* `zkv-node.Dockerfile`: generate a solo node image with a fresh source compilation (leverage on docker layers to create a small docker image)
+* `zkv-docker-compose.yaml`: the cluster definition that runs
```
ysibirski, line 82: Aren't you deleting this file as well?

ysibirski (changes requested):
I think you are missing more code cleanups under WORKFLOWS where we build docker images for solo and relay node binaries. Take a look at CI-tag-orchestrator 

This includes:
- CI-tag-orchestrator 
- actions/build-docker-image
- CI-build-docker-image

At the same time we may want to review the naming convention since relay docker builds we were naming `...-relay`. Since we are not building solo anymore perhaps we can drop `-relay` for the tags as well :thinking: 
-

docker/README.md
drgora: Yes indeed, messed up with conflicts on rebase, thanks for cathing! Removed

.github/workflows/CI-try-runtime.yml
drgora: Yes this is intended; this step will fail until the RPC node is connected to the relay chain.
The plan is to switch it back to `yes` again once that will be done.

DanieleDiBenedetto:
> I think you are missing more code cleanups under WORKFLOWS where we build docker images for solo and relay node binaries. Take a look at CI-tag-orchestrator
> 
> This includes:
> 
> * CI-tag-orchestrator
> * actions/build-docker-image
> * CI-build-docker-image
> 
> At the same time we may want to review the naming convention since relay docker builds we were naming `...-relay`. Since we are not building solo anymore perhaps we can drop `-relay` for the tags as well 🤔

I would avoid too much work on renaming on both our sides, calling it relay is fine for me

paratest/node/src/service.rs
```diff
@@ -217,11 +219,10 @@ async fn start_node_impl<Network: sc_network::NetworkBackend<Block, Hash>>(
         let client = client.clone();
         let transaction_pool = transaction_pool.clone();
 
-        Box::new(move |deny_unsafe, _| {
+        Box::new(move |_| {
```
DanieleDiBenedetto, line 222: Is something changed also in the way we specify if to allow/disallow unsafe RPC methods when running an RPC node ?

relay-node/benchmarks/src/reference_hardware.json
```diff
@@ -13,10 +13,10 @@
     },
     {
         "metric": "DiskSeqWrite",
-        "minimum": 536.40
+        "minimum": 220.0
```
DanieleDiBenedetto, line 16: Why this big difference ?

drgora: That's a great question, there is something weird indeed. The machine benchmarks on our reference machine were failing on these metric, so I set them accordingly. But this is probably unexpected and undesirable.
Maybe we should revert this change for this PR, and investigate further? cc @la10736

.github/workflows/CI-try-runtime.yml
ysibirski: If this is temp you can use the switch when calling the workflow vs changing its default config. For example 
```
uses: ./.github/workflows/CI-try-runtime.yml
    with:
      execute-try-runtime: 'no'
```

ysibirski:
> > I think you are missing more code cleanups under WORKFLOWS where we build docker images for solo and relay node binaries. Take a look at CI-tag-orchestrator
> > This includes:
> > 
> > * CI-tag-orchestrator
> > * actions/build-docker-image
> > * CI-build-docker-image
> > 
> > At the same time we may want to review the naming convention since relay docker builds we were naming `...-relay`. Since we are not building solo anymore perhaps we can drop `-relay` for the tags as well 🤔
> 
> I would avoid too much work on renaming on both our sides, calling it relay is fine for me

Sounds good regarding leaving the names, but cleanups are still needed since building docker image workflows will fail if we do not delete logic that builds solo binary image :wink:

paratest/node/src/service.rs
drgora: Yes; `deny_unsafe` is specific to an rpc interface now (and there can be many different interfaces).
The CLI switch to configure custom interfaces is `--experimental-rpc-endpoint`

la10736 (changes requested):
Great job, just some minor things. The most valuable one is about the use of `CARGO_WORSPACE_DIR` env.

DanieleDiBenedetto (changes requested):
LGTM ! Just a few comments. Thanks very much for the massive work !

pallets/verifiers/Cargo.toml
```diff
@@ -32,10 +32,12 @@ hp-on-proof-verified = { workspace = true }
 hp-verifiers = { workspace = true }
 pallet-verifiers-macros = { workspace = true }
 
+hex-literal = { workspace = true }
+
```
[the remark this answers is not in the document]
drgora: Yes, good catch, thanks

.cargo/config.toml
```diff
@@ -1,3 +1,6 @@
 [target.x86_64-unknown-linux-gnu]
 linker = "/usr/bin/cc"
 rustflags = ["-C", "link-arg=-fuse-ld=lld"]
+
+[env]
+CARGO_WORKSPACE_DIR = { value = "", relative = true }
```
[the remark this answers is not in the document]
drgora: Reverted the change

runtime/src/lib.rs
```diff
@@ -398,31 +405,6 @@ impl pallet_balances::Config for Runtime {
...
-        pub babe: Babe,
-        pub grandpa: Grandpa,
-        pub im_online: ImOnlineId,
-    }
-}
-
-#[cfg(feature = "relay")]
 impl_opaque_keys! {
     pub struct SessionKeysRelay {
         pub babe: Babe,
```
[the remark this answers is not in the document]
drgora: Makes sense, done

runtime/src/lib.rs
```diff
@@ -718,19 +716,6 @@ static_assertions::const_assert!(MIN_PROOFS_FOR_ROOT_PUBLISHING > 0);
 // We should keep in memory at least one attestation
 static_assertions::const_assert!(MAX_STORAGE_ATTESTATIONS > 1);
```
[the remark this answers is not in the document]
drgora: Right, removed all the unused constants

utils/native-cache/src/lib.rs
```diff
@@ -105,6 +102,7 @@ pub fn handle_dependencies<'a>(
     if skip_native_cache() {
         return Ok(());
     }
+    println!("Handling {:?}", target_root.as_ref());
```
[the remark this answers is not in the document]
drgora: Indeed, removed

runtime/src/payout.rs
```diff
@@ -69,16 +80,16 @@ parameter_types! {
     /// The target precions for exp(); impacts on the final precision for inflation computation.
     pub const ExpPrecision: f64 = 10e-15f64;
     /// Base inflation (I_b).
-    pub InflationBase: Perquintill = Perquintill::from_percent(5);
+    pub InflationBase: Perquintill = Perquintill::from_percent(3);
```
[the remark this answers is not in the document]
drgora: Done

DanieleDiBenedetto (approved):

ysibirski (approved):
LGTM

la10736 (approved):
LGTM

PR #246, opened 2025-03-11
feat: Add pallet token gateway + pallet hyperbridge

rdoria1:


runtime/src/pallet_assets_mock.rs
```diff
@@ -0,0 +1,244 @@
+// Mock implementation for Assets that satisfies token_gateway requirements
```
DanieleDiBenedetto, line 1: `pallet_token_gateway` `Assets` config parameter requires only an implementer of `fungible::Mutate` and `fungible::metadata::Mutate` so you don't necessarily need to "replicate" the structure of a Pallet here

DanieleDiBenedetto (changes requested):

runtime/src/lib.rs
```diff
@@ -1076,6 +1078,44 @@ impl ismp_grandpa::Config for Runtime {
...
+
+impl pallet_assets_mock::Config for Runtime {
+    type Currency = Balances;
+    type AssetId = u32;
+    type Balance = Balance;
+}
+
+parameter_types! {
+    pub const Decimals: u8 = 18;
+    pub const NativeAssetId: u32 = 0;
```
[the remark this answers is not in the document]
DanieleDiBenedetto: About Decimals I don't know if we can take it from somewhere else, probably not, but maybe worth spending (just) 5 minutes on it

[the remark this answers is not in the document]
rdoria1: > Take a look at the parameter `FeeAssetId` in `runtime/src/xcm_config.rs` and the way it's built. We could directly use that parameter instead (just for the sake of being consistent everywhere with our AssetId)

It's not the same AssetId, the one you are refering to is from staging xcm. The XCM's AssetId is a different type entirely - it's designed for cross-chain asset identification and uses a more complex structure involving Location. The one here is from pallet token gateway and is a different type.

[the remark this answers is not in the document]
rdoria1: > About Decimals I don't know if we can take it from somewhere else, probably not, but maybe worth spending (just) 5 minutes on it

We don't, we do have it as internal var from the specs but we can't really import that unfortunately. https://github.com/zkVerify/zkVerify/blob/7c86a7a2b5dd41768fcbf04737d70ad6c157f638/relay-node/service/src/chain_spec.rs#L51

runtime/src/pallet_assets_mock.rs
rdoria1: true I removed a few things but kept the minimal needed for it to compile

DanieleDiBenedetto (approved):

runtime/src/lib.rs
```diff
@@ -1007,7 +1009,7 @@ impl pallet_ismp::Config for Runtime {
 
 impl pallet_hyperbridge_aggregations::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
-    type IsmpDispatcher = pallet_ismp::Pallet<Runtime>;
+    type IsmpDispatcher = pallet_hyperbridge::Pallet<Runtime>;
```
lgiussan, line 1012: Is this change needed? What are the consequences?

runtime/src/pallet_assets_mock.rs
```diff
@@ -0,0 +1,190 @@
...
```
drgora, line 1: It seems to me that this mock in never actually used in the current configuration, not even for the benchmarks.

If this is correct, I would suggest replacing all the random implementations of this pallet with something like `unimplemented!`, with some sensible message to make it clear that it is not supposed to be used.

runtime/src/lib.rs
```diff
@@ -1017,6 +1019,40 @@ impl ismp_grandpa::Config for Runtime {
...
+    pub AssetAdmin: AccountId = Treasury::account_id();
+}
+
+impl pallet_token_gateway::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type Dispatcher = Ismp;
+    type Assets = pallet_assets_mock::Pallet<Runtime>;
+    type NativeCurrency = Balances;
+    type AssetAdmin = AssetAdmin;
+    type CreateOrigin = EnsureRoot<AccountId>;
```
drgora, line 1049: It seems that this breaks the automatic benchmarks for us; given that these benchmarks do not use any `try_successful_origin()`, I would suggest using a different `CreateOrigin` when compiling with the feature `runtime-benchmarks`?

drgora (changes requested):
Looks mostly good to me, but I think we should fix, or at least handle explicitly, the benchmarks before merging.

runtime/src/lib.rs
rdoria1: yes this change is needed, Actually the ISMP dispatcher will be used to dispatch cross-chain requests. Pallet hyperbridge provides an IsmpDispatcher implementation that handles both protocol and relayer fee collection.
So actually this is the way to say that we are actually using pallet hyperbridge in the whole flow and fee will be collected.

drgora (approved):
LGTM, thanks!

DanieleDiBenedetto:
@rdoria1 shouldn't we re-bench pallet_aggregate as well considering that now is not using anymore pallet_ismp for bridging ?

rdoria1:
> @rdoria1 shouldn't we re-bench `pallet_aggregate` as well considering that now is not using anymore `pallet_ismp` directly ?

Very good point, yes indeed it would be better to do it as well

PR #250, opened 2025-03-17
New genesis base

la10736:
Here we introduce a the new genesis: that's just a draft where following steps are missing

1. Generate the final key and use the new public key to file the genesis
2. Create the docker image and generate the genesis in raw mode
3.  Replace the file

Beside that, we did little improving in the script for dump and restore balance:
- move the filtering from dump to store
- filter all balance under the existential deposit
- filter all modules account
- cap account balance (default 1000)
- add option to define the chain address
- other little stuffs


relay-node/service/src/chain_spec.rs
```diff
@@ -47,7 +47,7 @@ fn chain_properties() -> Properties {
             "ss58Format".to_string(),
             serde_json::Value::from(zkv_runtime::SS58Prefix::get()),
         ),
-        ("tokenSymbol".to_string(), serde_json::Value::from("ACME")),
+        ("tokenSymbol".to_string(), serde_json::Value::from("VFY")),
```
DanieleDiBenedetto, line 50: For testnet it's `tVFY`

runtime/build.rs
```diff
@@ -24,7 +24,7 @@ fn main() {
         // variable.
         #[cfg(feature = "metadata-hash")]
         let builder = if std::env::var_os("ZKV_FORCE_DISABLE_METADATA_HASH").is_none() {
-            builder.enable_metadata_hash("ACME", 18)
+            builder.enable_metadata_hash("VFY", 18)
```
DanieleDiBenedetto, line 27: For testnet it's `tVFY`

scripts/generate_key.py
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+
+def generate_secret_phrase(base_key, name_component, index):
+    if base_key is None:
+        return None
+    return f"{base_key}///{name_component}/{index}"
+
+
+def generate_validator_keys(key_generator_path, num_validators, base_key=None, separate_account=False):
```
DanieleDiBenedetto, line 30: Minor, non-blocker: Can't we re-use here the `generate_sr25519_keys` method ? Perhaps for better readability we can introduce also a `generate_ed25519_keys` method ?

scripts/generate_key.py
```diff
@@ -0,0 +1,210 @@
...
+        keys.append({
+            'sr25519_public_key': data['ss58_public_key']
+        })
+
+        secrets.append({
+            'secret_phrase': sanitize_secret_phrase(base_key, secret_phrase)
+        })
+    return keys, secrets
+
+
```
DanieleDiBenedetto, line 122: Add some doc on usage ?

DanieleDiBenedetto (changes requested):

scripts/generate_key.py
la10736: Should be already done by `argparse`

scripts/snapshot-restore-balance/filter_account.json
```diff
@@ -0,0 +1,14 @@
+[
```
DanieleDiBenedetto, line 1: I don't think we want to filter any account for now, considering we have re-designed the token allocation

runtime/src/genesis_config_presets.rs
```diff
@@ -250,30 +229,408 @@ fn default_parachains_host_configuration(
...
+        ValidatorData::from_ids(
+            "5Cd7HLhDtvzUsmsmBZ7tvGHCdtCofEiAfcVnxWW9U1SyaU9C",
+            "5Cd7HLhDtvzUsmsmBZ7tvGHCdtCofEiAfcVnxWW9U1SyaU9C",
+            "5Euet7r4REQogSfNwBjqvwweFPq7hHsTgaU91dZf2xG9eskf",
+            VALIDATOR_BALANCE,
+            VALIDATOR_BOND,
+        )?,
+    ];
+
+    let nominators = [
```
DanieleDiBenedetto, line 480: We don't need an initial nominator set for now

runtime/src/genesis_config_presets.rs
```diff
@@ -250,30 +229,408 @@ fn default_parachains_host_configuration(
...
+            "5FZboFp74WMi3ogERtYbmLwiZmbhTbxhfApAmf4jSgdw98kR",
+            INVESTOR_BALANCE,
+        )?,
+        FoundedAccount::from_id(
+            "5FTyReguLB6VN2iWzv9ZqHfM8ZtsE9CPiXyWS89A3xaheEF7",
+            INVESTOR_BALANCE,
+        )?,
+    ];
+
+    let initial_authorities = [
```
DanieleDiBenedetto, line 407: Only 7 validators to start

DanieleDiBenedetto (changes requested):

runtime/src/genesis_config_presets.rs
la10736: Done

runtime/src/genesis_config_presets.rs
la10736: Done

DanieleDiBenedetto (approved):
LGTM !

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -72,28 +103,81 @@ async function restoreBalances(api, keyring, snapshotBalancesFile, custodySeed)
     }
 }
 
+function usage() {
+    console.error(`Usage : snapshot or restore
+    node snapshot-balances.js [-e ] snapshot
```
drgora, line 108: I'm afraid that having only the `-e` option might be a bit confusing. I'd suggest to either have all the options here, or none.

scripts/snapshot-restore-balance/snapshot-balances.js
```diff
@@ -72,28 +103,81 @@ async function restoreBalances(api, keyring, snapshotBalancesFile, custodySeed)
...
+        or 
+    node snapshot-balances.js restore "//Alice"
```
drgora, line 110: How about something like the following?
```suggestion
    node snapshot-balances.js restore "//CustodyWalletSeed"
```

runtime/src/genesis_config_presets.rs
```diff
@@ -93,78 +94,57 @@ pub fn authority_keys_from_seed(
     )
 }
 
-/*
 // Generate authority IDs from SS58 addresses.
 pub fn authority_ids_from_ss58(
-    sr25519_key: &str,
+    sr25519_account_key: &str,
+    sr25519_common_key: &str,
     ed25519_key: &str,
```
drgora, line 101: Using a name like `session_key` might make the meaning of this key more clear?
```suggestion
    sr25519_session_key: &str,
    ed25519_session_key: &str,
```

drgora (approved):
Looks good to me! I just left few non-blocking comments

scripts/snapshot-restore-balance/snapshot-balances.js
la10736: It's a leftover... thanks!

PR #266, opened 2025-03-25
Refactor reference_machine_run_all.sh:

cronicc:
* Use Linux brd (block ram disk) driver instead of loop device for benchmark disk

* Format disk with FAT32 which supports mounting with '-o sync' forcing O_DIRECT disk access

  * This makes cgroup.blkio namesspace limits work in docker consistently

* Disable swap prior to running the benchmark, reenable after

* Limit the number of CPUs to 8 vCPUs

  * Mirrors the CPU topology of an AWS c7a.2xlarge instance

  * Weights of pallets benefiting from multithreading will slightly increase

* Run 'run_all_benchmarks.sh' once without any enabled benchmarks before the actual test

  * This will trigger a docker build (if needed) before underclocking the CPU which will be faster

* Improve error handling and restoring of the machine to the base settings on exit, error or interruption

* Disallow concurrent script execution through exclusive locking

### DIFF
--- a/scripts/reference_machine_run_all.sh
+++ b/scripts/reference_machine_run_all.sh
@@ -7,7 +7,7 @@ ECHO_CMD="${ECHO_CMD:-false}"
 
 # functions
 
-check_root () {
+check_root() {
   local extra_msg="$1"
   if [ "$(whoami)" != 'root' ]; then
     echo "$(date --utc +%FT%T.%3NZ) Error: Run this script with 'sudo $0'.${extra_msg}"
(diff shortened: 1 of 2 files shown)


scripts/reference_machine_run_all.sh
```diff
@@ -73,29 +73,70 @@ restore_cpu () {
...
+  local mountpoint="$1"
+  local USER_ID="$2"
+  local GROUP_ID="$3"
+  DEVICE="/dev/ram0"
+  # rd_size in KiB, allocate 1GiB
+  modprobe brd rd_nr=1 rd_size="$((1024**2*1))" max_part=1 && BRD_CREATED="true"
+  # fill with 1GiB of random data
+  dd if=/dev/urandom of="${DEVICE}" bs=1024 count="$((1024**2*1))"
+  # use FAT32 because it supports -o sync, logical sector size == brd physical sector size == page size == 4096
+  mkfs.vfat -F32 -S4096 "${DEVICE}"
```
DanieleDiBenedetto, line 87: Aren't we losing some performance on disk with FAT32 ?

DanieleDiBenedetto (approved):

scripts/reference_machine_run_all.sh
cronicc: The throughput and iops values in the cgroup settings have been adjusted to match the previous performance values in spite of using FAT32. The machine benchmark sequential write / random write values are within +/- 1% of previous values and now much more consistent. The point of limiting disk throughput and using RAM backed storage was to get consistent and repeatable numbers. This unfortunately did not work with EXT4 as cgroup limits only work for synchronous / O_DIRECT io. (Read more here https://andrestc.com/post/cgroups-io/)

la10736 (approved):
LGTM

PR #270, opened 2025-03-28
Filter out submit_proof extrinsics from Nontransfer proxy and related unit test.

la10736:
Filter out submit_proof extrinsics from Nontransfer proxy and related unit test
Also add plonky2 to nontransfer proxy conf.

Just to give a little context about why verifier's `submit_proof` extrinsics cannot be classified as a `NonTransfer`. The short answer is that `submit_proof` hold funds and, when somebody call `aggregate` **transfer**  this funds from the held ones to the delivery owner (also to the aggregator caller but for this issue that's not the main concern). Starting from this consideration it's simple to imagine some kind of privilege escalation where an account A define a `NonTransfer` proxy account B, now B can first create a fake domain, set a delivery price and then send a proof to this domain in behave of A to just stole the founds from A.

### DIFF
--- a/runtime/src/proxy.rs
+++ b/runtime/src/proxy.rs
@@ -33,13 +33,32 @@ impl Default for ProxyType {
         Self::Any
     }
 }
+
+impl ProxyType {
+    fn is_a_submit_proof_extrinsic(c: &RuntimeCall) -> bool {
+        matches!(
+            c,
+            RuntimeCall::SettlementGroth16Pallet(pallet_verifiers::Call::submit_proof { .. })
+                | RuntimeCall::SettlementRisc0Pallet(pallet_verifiers::Call::submit_proof { .. })
+                | RuntimeCall::SettlementUltraplonkPallet(
+                    pallet_verifiers::Call::submit_proof { .. }
+                )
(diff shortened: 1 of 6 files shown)


runtime/src/proxy.rs
```diff
@@ -65,9 +84,10 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
                 RuntimeCall::SettlementGroth16Pallet(..) |
                 RuntimeCall::SettlementRisc0Pallet(..) |
                 RuntimeCall::SettlementUltraplonkPallet(..) |
-                RuntimeCall::SettlementPlonky2Pallet(..) |
-                RuntimeCall::SettlementProofOfSqlPallet(..)
-            ),
+                RuntimeCall::SettlementProofOfSqlPallet(..) |
+                RuntimeCall::SettlementPlonky2Pallet(..)
```
drgora, line 88: Maybe we could add a one line comment to remind that also `is_a_submit_proof_extrinsic` should be updated when adding new verifiers here?

drgora (approved):
LGTM

DanieleDiBenedetto (approved):

PR #278, opened 2025-04-15
Re-introduce FFLONK verifier

95DDB:
Also introduced the vk hash API for FFLONK

### DIFF
--- a/rpc/vk_hash/src/lib.rs
+++ b/rpc/vk_hash/src/lib.rs
@@ -17,7 +17,7 @@ use codec::{Decode, Encode};
 use hp_verifiers::Verifier;
 use jsonrpsee::{core::RpcResult, proc_macros::rpc, types::ErrorObject};
 use pallet_fflonk_verifier::{
-    vk::{Fq, Fq2, Fr, Vk, G1, G2},
+    vk::{Fq, Fq2, Fr, G1, G2},
     Fflonk,
 };
 use pallet_groth16_verifier::{Curve, Groth16};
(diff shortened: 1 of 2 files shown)


zombienet-tests/js_scripts/0014-vk_hash_api.js
DimitrisPapac, line 1: Since `PROOF` and `PUBS` are not used at all within this particular zombie-net test's logic, here, I would suggest that we omit the `PROOF` and `PUBS` entries in each verifier dictionary that we define.

rpc/vk_hash/src/lib.rs
```diff
@@ -83,6 +103,26 @@ impl VKHash {
 }
 
 impl VKHashApiServer<H256> for VKHash {
+    fn fflonk(&self, vk: FflonkVk) -> RpcResult<H256> {
+        let vk = Vk {
```
la10736, line 107: Implement `From<FflonkVk> for pallet_fflonk_verifier::Vk` and use `into()` is nicer and idiomatic (also for groth16).

runtime/src/lib.rs
```diff
@@ -892,6 +892,16 @@ impl pallet_verifiers::common::Config for Runtime {
...
+impl pallet_verifiers::Config<pallet_fflonk_verifier::Fflonk> for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type OnProofVerified = Aggregate;
+    type WeightInfo =
+        pallet_fflonk_verifier::FflonkWeight<weights::pallet_fflonk_verifier::ZKVWeight<Runtime>>;
+    type Ticket = VkRegistrationHoldConsideration;
+    #[cfg(feature = "runtime-benchmarks")]
+    type Currency = Balances;
+}
+
```
la10736, line 904: Just some thoughts. Maybe is worth of trying to refactor a bit and move all configuration that can be moved in `common::Config`. I guess for sure `Currency`, `Ticket`. For `RuntimeEvent` I'm not sure and `OnProofVerifier` maybe is possible to implement `()` that use the one in `common::Config`, an override and accumulation.

Let me know if I can spend some hours on this.

zombienet-tests/zkv-lib/index.js
la10736, line 1: Why the format is changed? You'd run the format on your side and on our side did vscode use a wrong one?

We need to double check it in order to not start a format dance.

la10736 (commented):
No blockers  but we should check the js format conventions.

runtime/src/lib.rs
DanieleDiBenedetto: I'll open a ticket thanks

zombienet-tests/js_scripts/0014-vk_hash_api.js
DanieleDiBenedetto: Done

rpc/vk_hash/src/lib.rs
DanieleDiBenedetto: Done

zombienet-tests/zkv-lib/index.js
DanieleDiBenedetto: Will be addressed in another PR, sorry

DimitrisPapac (approved):

la10736 (approved):
LGTM

PR #284, opened 2025-04-28
Bump runtime version and add test to check coherence between spec_version and crate version

la10736:


runtime/src/tests/misc.rs
```diff
@@ -103,3 +103,19 @@ fn submit_proof_weights_composition_should_ignore_aggregate_if_no_domain() {
...
+    let v_num = v
+        .split('.')
+        .map(|x| x.parse::<u32>().unwrap())
+        .enumerate()
+        .fold(0, |a, (step, dec)| a + dec * 1000_u32.pow(step as u32));
+    use sp_api::runtime_decl_for_core::CoreV5;
+    let s_ver = Runtime::version().spec_version;
+    assert_eq!(
+        s_ver, v_num,
+        "Version missmatch. Crate version = {v}, but spec_version is {s_ver} != {v_num}"
```
drgora, line 119:
```suggestion
        "Version mismatch. Crate version = {v}, but spec_version is {s_ver} != {v_num}"
```

runtime/src/tests/misc.rs
```diff
@@ -103,3 +103,19 @@ fn submit_proof_weights_composition_should_ignore_aggregate_if_no_domain() {
...
 }
+
+#[test]
+fn check_version() {
+    let v = std::env!("CARGO_PKG_VERSION");
+    let v_num = v
+        .split('.')
+        .map(|x| x.parse::<u32>().unwrap())
+        .enumerate()
+        .fold(0, |a, (step, dec)| a + dec * 1000_u32.pow(step as u32));
```
drgora, line 114: Shouldn't we reverse before `enumerate()`?
Otherwise, for example, a crate version `0.14.1` would result in `v_num == 1014000`, while it should be `14001`.

drgora (changes requested):

DanieleDiBenedetto (approved):

PR #285, opened 2025-04-28
Substrate parameter value true release for usage

ysibirski:
Some of the substrate config parameters are using actual value `true`. For example: `--enable-offchain-indexing true`

Therefore this PR enables `true` as a possible value for a parameter and for all the rest like --validator using `yes` instead, that can easily be changed and set under .env file if needed.


lander86 (approved):
LGTM

AndrewCoal737 (approved):

la10736 (approved):
LGTM

drgora (approved):

PR #287, opened 2025-05-09
Fix CI 2025-05-09

la10736:
1. Update sp-io from 38.0.0 to 38.0.2 to fix a new issue with last rust nightly version (1.88)
2. Add two audit exceptions relate an old version of wasmtime that cannot be updated (transitive dependency)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3825,7 +3825,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "18e4fdb82bd54a12e42fb58a800dcae6b9e13982238ce2296dc3570b92148e1f"
 dependencies = [
  "data-encoding",
-    "syn 2.0.100",
+ "syn 2.0.100",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


DanieleDiBenedetto (approved):

Cargo.lock
lgiussan, line 1: For some reason, indentation went from 1 space to 4 spaces in the changed lines in the Cargo.lock. I would suggest to check your development environment for why this is happening and commit an updated Cargo.lock with 1 space indentation.

lgiussan (changes requested):

Cargo.lock
la10736: Ok, the strange thing is that it was changed by `cargo update`. Later I opened it on my IDE and I saw that it try to commit back the 1 space format but I discarded it because I though that now was wrong... :thinking:

la10736: Fixed, it was a rustrover configuration that in the first commit applied it to all changed lines

lgiussan (approved):
LGTM, thank you

PR #288, opened 2025-05-09
Update XCM config

drgora:
This updates the XCM config so that:
- teleports are enabled for the zkVerify-EVM-Parachain (paraId `1`)
- xcm execution fees are computed and handled like all the other fees
- the networkId is updated to represent the Volta network

### DIFF
--- a/runtime/src/xcm_config.rs
+++ b/runtime/src/xcm_config.rs
@@ -50,7 +50,7 @@ use crate::weights::pallet_xcm::ZKVWeight as XcmPalletZKVWeight;
 use crate::weights::xcm::ZKVWeight as XcmZKVWeight;
 
 const ZKV_GENESIS_HASH: [u8; 32] =
-    hex_literal::hex!("e2a4f521dbcba897cd2359adc5e7725f409b17f9ae129737904e5e8939c22a05");
+    hex_literal::hex!("ff7fe5a610f15fe7a0c52f94f86313fb7db7d3786e7f8acf2b66c11d5be7c242");
 
 parameter_types! {
     pub const RootLocation: Location = Here.into_location();
@@ -138,16 +138,25 @@ pub type XcmRouter = WithUniqueTopic<(
     ChildParachainRouter<Runtime, XcmPallet, PriceForChildParachainDelivery>,
 )>;
 
+// zkVerify-EVM-Parachain
+pub const ZKV_EVM_PARA_ID: u32 = 1;
+// paratest, included in this repository
 pub const TEST_PARA_ID: u32 = 1599;
+
 parameter_types! {
-    pub const Acme: AssetFilter = Wild(AllOf { fun: WildFungible, id: AssetId(TokenLocation::get()) });
+    pub const TVFYAsset: AssetFilter = Wild(AllOf { fun: WildFungible, id: AssetId(TokenLocation::get()) });
     pub TestParaLocation: Location = Parachain(TEST_PARA_ID).into_location();
-    pub AcmeForTest: (AssetFilter, Location) = (Acme::get(), TestParaLocation::get());
+    pub ZKVEvmParaLocation: Location = Parachain(ZKV_EVM_PARA_ID).into_location();
+    pub TVFYAssetForTest: (AssetFilter, Location) = (TVFYAsset::get(), TestParaLocation::get());
+    pub TVFYAssetForZKVEvm: (AssetFilter, Location) = (TVFYAsset::get(), ZKVEvmParaLocation::get());
     pub const MaxAssetsIntoHolding: u32 = 64;
 }
 
 /// ZKV Relay recognizes/respects Test parachain as teleporter for VFY.
-pub type TrustedTeleporters = xcm_builder::Case<AcmeForTest>;
+pub type TrustedTeleporters = (
+    xcm_builder::Case<TVFYAssetForTest>,
+    xcm_builder::Case<TVFYAssetForZKVEvm>,
+);
 
 pub struct OnlyParachains;
 impl Contains<Location> for OnlyParachains {
--- a/runtime/src/xcm_config.rs
+++ b/runtime/src/xcm_config.rs
@@ -21,19 +21,15 @@ use super::{
     AccountId, AllPalletsWithSystem, Balances, Dmp, ParaId, Runtime, RuntimeCall, RuntimeEvent,
     RuntimeOrigin, TransactionByteFee, XcmPallet,
 };
-use crate::parachains::parachains_origin;
 use frame_support::{
     parameter_types,
     traits::{Contains, Equals, Everything, Nothing},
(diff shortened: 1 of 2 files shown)


runtime/src/xcm_config.rs
```diff
@@ -50,7 +46,7 @@ use crate::weights::pallet_xcm::ZKVWeight as XcmPalletZKVWeight;
 use crate::weights::xcm::ZKVWeight as XcmZKVWeight;
 
 const ZKV_GENESIS_HASH: [u8; 32] =
-    hex_literal::hex!("e2a4f521dbcba897cd2359adc5e7725f409b17f9ae129737904e5e8939c22a05");
+    hex_literal::hex!("ff7fe5a610f15fe7a0c52f94f86313fb7db7d3786e7f8acf2b66c11d5be7c242");
```
DanieleDiBenedetto, line 49: This is the current genesis hash right ? The old one was outdated ?

runtime/src/xcm_config.rs
```diff
@@ -138,16 +134,25 @@ pub type XcmRouter = WithUniqueTopic<(
...
 
+// zkVerify-EVM-Parachain
+pub const ZKV_EVM_PARA_ID: u32 = 1;
+// paratest, included in this repository
 pub const TEST_PARA_ID: u32 = 1599;
+
 parameter_types! {
-    pub const Acme: AssetFilter = Wild(AllOf { fun: WildFungible, id: AssetId(TokenLocation::get()) });
+    pub const TVFYAsset: AssetFilter = Wild(AllOf { fun: WildFungible, id: AssetId(TokenLocation::get()) });
     pub TestParaLocation: Location = Parachain(TEST_PARA_ID).into_location();
```
DanieleDiBenedetto, line 144: Is there a way around the fact that we are keeping here stuff related to the Para testing ?

runtime/src/xcm_config.rs
```diff
@@ -138,16 +134,25 @@ pub type XcmRouter = WithUniqueTopic<(
...
-    pub AcmeForTest: (AssetFilter, Location) = (Acme::get(), TestParaLocation::get());
-    pub const MaxAssetsIntoHolding: u32 = 64;
+    pub ZKVEvmParaLocation: Location = Parachain(ZKV_EVM_PARA_ID).into_location();
+    pub TVFYAssetForTest: (AssetFilter, Location) = (TVFYAsset::get(), TestParaLocation::get());
+    pub TVFYAssetForZKVEvm: (AssetFilter, Location) = (TVFYAsset::get(), ZKVEvmParaLocation::get());
+    pub const MaxAssetsIntoHolding: u32 = 1;
```
DanieleDiBenedetto, line 148: Polkadot sets this to 64. In which occasions the Assets are getting trapped ?

runtime/src/xcm_config.rs
```diff
@@ -138,16 +134,25 @@ pub type XcmRouter = WithUniqueTopic<(
...
 )>;
 
+// zkVerify-EVM-Parachain
+pub const ZKV_EVM_PARA_ID: u32 = 1;
+// paratest, included in this repository
 pub const TEST_PARA_ID: u32 = 1599;
+
 parameter_types! {
-    pub const Acme: AssetFilter = Wild(AllOf { fun: WildFungible, id: AssetId(TokenLocation::get()) });
+    pub const TVFYAsset: AssetFilter = Wild(AllOf { fun: WildFungible, id: AssetId(TokenLocation::get()) });
```
DanieleDiBenedetto, line 143: Let's simply call it tVFY ?

DanieleDiBenedetto (commented):
I was looking at other parameters in the xcm_config file:
- `BaseXcmWeight`, `BaseDeliveryFee`, `xcm_executor::Config::PriceForChildParachainDelivery`: shall we keep the same ?
- `xcm_executor::Config::FeeManager`: it seems we are sending xcm fee to treasury account, which is compatible with Polkadot. Question: Are these fees the one paid ONLY for the XCM execution on the relay chain ? If yes, shouldn't the validators of the relay chain, theoretically, receive them ? 
- I think we can remove lines 239 to 250 (related to governace/treasury body ids)
-

runtime/src/xcm_config.rs
drgora: Yes, this is Volta's. The previous genesis referred to the pre-regenesis chain.

runtime/src/xcm_config.rs
drgora: Yeah we could do it (similar to Polkadot), but I don't quite like the fact that the name would overlap with our constants in the `currency` module.
Maybe we should also rename that, for example `VFY` -> `UNITS`, also similar to Polkadot's approach?

runtime/src/xcm_config.rs
drgora: Not sure, the paratest must be whitelisted from the relay point of view, so the info must be available somewhere in the relay chain code.

Maybe we can change our Paratest id to `1` as well, and pretend that Paratest is the same as the evm para? But honestly I don't like that route

runtime/src/xcm_config.rs
drgora: This is not strictly related to assets being trapped. Given the simple scenarios that we want to support initially, and given that we do not plan to have any assets other than the native token for the time being, I thought there was no point in having more than `1` here. This is not a strong point though, so if we want to stick to Polkadot config, and be ready for unexpected uses of xcm, it's fine by me.

drgora:
>     * `BaseXcmWeight`, `BaseDeliveryFee`, `xcm_executor::Config::PriceForChildParachainDelivery`: shall we keep the same ? 

These are all great question, which deserve proper discussion in my opinion. But maybe the tokenomics involving xcm goes out of the scope of this PR?

>     * `xcm_executor::Config::FeeManager`: it seems we are sending xcm fee to treasury account, which is compatible with Polkadot. Question: Are these fees the one paid ONLY for the XCM execution on the relay chain ? If yes, shouldn't the validators of the relay chain, theoretically, receive them ?

These fees are not really for the execution of the xcm itself; the execution of the xcm extrinsic is computed and charged as any other extrinsic (i.e. with our own `DealWithFees` impl). These are extra fees incurred by some xcm specific operations, such as sending a message to another chain; I believe the idea is that these operations do not represent an execution cost for the block authors only, but rather involve several actors, possibly offchain.

>     * I think we can remove lines 239 to 250 (related to governace/treasury body ids)

Good idea, done

runtime/src/xcm_config.rs
DanieleDiBenedetto: I guess that's why polkadot named it using this case. As you wish, not a blocker, we can keep it like this

la10736 (approved):
LGTM

DanieleDiBenedetto (approved):

PR #307, opened 2025-06-25
Implement the possibility to clear all caches. 

la10736:
There are some cases where we NEED to remove all caches and rebuild all from scratch.

### DIFF
--- a/.github/workflows/CI-orchestrator.yml
+++ b/.github/workflows/CI-orchestrator.yml
@@ -72,7 +72,7 @@ jobs:
         feature-propagation-job,
         udeps-job,
       ]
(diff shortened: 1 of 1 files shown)


DanieleDiBenedetto (approved):

la10736:
> Just keep in mind that `build-runtime-benchmark` cache key was only merged into _ci_testing repo and does not exist here yet 😉
> 
> [zkVerify/zkVerify_ci_testing#37](https://github.com/zkVerify/zkVerify_ci_testing/pull/37)

Yes I know it, but try to remove something that dosn't exists yet should not be an issue ... right :smile:

PR #309, opened 2025-06-27
Configure Relay chain for EVM Parachain Support

95DDB:
Also:
- Prepared a test_parachain_host_configuration() which is going to be used from now on for all the tests. **NOTE: We have to manually update the chain with the new parameters !!!**
- Added CI step for executing benchmarks (as to check none breaks if we change runtime configuration at a certain point)
- Added cargo-make and make file
- Added format to zepter. This fixed the format in all tomls, so best if you filter out these files while reviewing


la10736 (changes requested):
Some little fixes in the `Makefile.toml` and maybe we can just update the default parachain host with the new paramenters.

drgora (approved):
LGTM

PR #310, opened 2025-06-30
Add a template to verify runtime code

la10736:


Cargo.toml
```diff
@@ -18,6 +18,7 @@ members = [
     "pallets/claim",
     "primitives/hp-verifiers",
     "primitives/hp-groth16",
+    "verifier-nostd-check",
```
DanieleDiBenedetto, line 21: Can we put this in the [exclude] section ? Otherwise it gets compiled every time.

DanieleDiBenedetto (changes requested):

lander86 (approved):

PR #311, opened 2025-06-30
Fix some warning about `mismatched_lifetime_syntaxes` that will land in stable in the next versions

la10736:


drgora (approved):
Thanks for the heads up! :)

PR #324, opened 2025-07-10
Add support for risc0 v2.2

lgiussan:


lgiussan:
> Perhaps we want to update also zombienet tests ?

Should the zombienet test be for the v2.2 only, or for both v2.1 and v2.2?

DanieleDiBenedetto (approved):

DimitrisPapac (approved):

PR #349, opened 2025-09-18
Add support for risc0 v3.0

lgiussan:


zombienet-tests/js_scripts/0009-proof_on_disabled_verifier_should_pay_just_a_little_fee.js
```diff
@@ -10,7 +10,7 @@ const ReturnCode = {
 };
 
 const { init_api, submitProof, getBalance, receivedEvents, submitExtrinsic } = require('zkv-lib')
-const { PROOF, PUBS, VK } = require('./risc0_v2_3_data.js');
```
la10736, line 13: I'm not sure I would remove this test: we have several proofs on testnet with 2.x version, so maybe is better maintain this test too

la10736: BTW, maybe is better to recover the 2.2 too till we have some proofs in testnet with this version.

PR #352, opened 2025-09-19
New Claim Pallet

DanieleDiBenedetto:
Main changes:
- On new claim, a message to be signed by beneficiaries should be passed
- Claim is now an unsigned extrinsic. A signature on the claim message with the beneficiary address must be submitted to successfully claim
- End claim will start removing up to T::MaxOpBeneficiaries
- If there are more left, manager should manually remove the remaining ones (TBD: leverage pallet hooks ?).
- If there are old claimants pending in storage it's not possible to initiate a new one


zombienet-tests/js_scripts/0015-claim_no_fee.js
```diff
@@ -0,0 +1,114 @@
...
+
+    // Check beneficiary balance
+    let balance_beneficiary = (await api.query.system.account(beneficiary_sr.address))["data"]["free"];
+    if (balance_beneficiary != '1000000000000000000') {
+        console.log(`Beneficiary SR balance is ${balance_beneficiary}, expected 1000000000000000000`);
+        return ReturnCode.ErrClaimedAmount;
+    }
+
+    // Claim SR Prefixed
+    signature = "0xc89703e8763b08dfedd8e78959d1c1b28138628e58c7e4bc2c0eefd87d796c10e3fdb7bd35ef713d02c49959d3bfb2c559673a70e54156bc71bff5e7541f1e86";;
```
drgora, line 67: It might be a good future reference to add a comment on how this signature was generated

drgora (changes requested):
Nice! I left few minor comments, with the only one that I believe needs attention before merging is about reverting the modifications to `runtime/zkverify/src/weights/block_weights.rs`

PR #354, opened 2025-09-19
Stwo verifier: fully functional + pallet events + tests

rhapy01:
All tests passing. Pallet events fixed. Added verify_batch and CLI wiring. Rebase-only policy: keep branch synced via rebase onto upstream/main; use --force-with-lease after green tests.

### DIFF
--- a/verifiers/stwo/src/stwo.rs
+++ b/verifiers/stwo/src/stwo.rs
@@ -222,8 +222,28 @@ impl StwoVerifier {
         let constraint_sum: u32 = proof.constraint_polynomials_lde_commitment.iter().map(|&x| x as u32).sum();
         let input_sum: u32 = public_inputs.inputs.iter().map(|&x| x as u32).sum();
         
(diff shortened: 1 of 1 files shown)


rhapy01:
Hi maintainers, could you please approve/rerun workflows for this PR? This is a first-time contribution from a fork, so CI checks are pending approval. All local tests pass (pallet-zkverify-starky 8/8, zkv-stwo 12/12). Thanks!

.github/workflows/verify.yml
ysibirski, line 1: The syntax of this workflow does not comply with our pattern.

- We do no want to trigger it on every new commit push to the opened PR.
- We use warpbuild builders vs `ubuntu-latest`
- We use composite action cmd-in-docker that is using docker image that comes with all the dependencies installed, so you do not need to run apt-get .... every time and it takes care of cargo cache

Take a look at the existing workflows for the reference

cc: @DanieleDiBenedetto @la10736 
-

PR #401, opened 2026-02-17
Update to upstream 2512

la10736:
Update dependencies to 2512.


95DDB (approved):

relay-node/service/upstream-differences.md
```diff
@@ -0,0 +1,373 @@
+# zkv-service vs upstream polkadot-service (polkadot-stable2512)
+
+This document summarizes the differences between `zkv-service` (at `relay-node/service/`) and
+the upstream `polkadot-service` crate from the `polkadot-stable2512` branch of `polkadot-sdk`.
```
drgora: Actually, I see that pretty much every md file references specific upstream versions.
Maybe we could add something like a `_2512` suffix to the file names, when they are not meant to be updated on the next release?

drgora (approved):
Looks good to me! I just left few non-blocking comments on some mds

drgora (approved):
Left few more non-blocking comments

docs/investigate_benchmark_issue.md
```diff
@@ -0,0 +1,31 @@
+# Overview
+
+In this branch we upgraded the code to `stable2512` polkadot branch. Some benchmarks do 
+not work. Some of them are new in this version and other could be regressions leaded by wrong 
+runtime configuration.
+
+## Commands
+
+* Compile: `cargo build -p zkv-relay --features runtime-benchmarks`
+* Given a <pallet> (could be a list) and <extrinsic> (could be a list) : `./target/release/zkv-relay benchmark pallet --runtime ./target/release/wbuild/zkv-runtime/zkv_runtime.compact.compressed.wasm --pallet <pallet> --extrinsic "<extrinsic>" -s 2 -r 1`
```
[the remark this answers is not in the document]
la10736: Right, I chose to use debug in 9 to  speed up the loop but later I forgot to point to debug... Fun fact: I usd thi doc with claude to fix some issues and he replaced the `release` with target without asking anything. Anyway, I'll move this docs in a SKILL.

docs/investigate_benchmark_issue.md
```diff
@@ -0,0 +1,31 @@
...
+* Understand if that benchmark was still present in the previous version (check it against main branch)
+* Understand how Polkadot, Kusama and Westend runtimes implements the configuration for this benchmarks
+* Compare with our implementation
+* Fix the issue
+* Provide issue description and solving path
+
+## Take care of
+
+* Only code in `runtime` folder should be changed and mostly should be gated by `runtime-benchmarks` feature.
+* The previous code was based on `stable2412` branch.
```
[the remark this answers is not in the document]
la10736: Yeah, I wrote this doc to solve some benchmarking issues that I had... after I realized that could be useful to have it here for further issues, but I didn't care to make it generic. I'll remove it in favor to a generic SKILL/COMMAND.

docs/refine_patchs.md
la10736, line 1: Also this doc can be translated in a SKILL

paratest/node/src/service.rs
```diff
@@ -13,7 +13,7 @@ use paratest_runtime::{
 // Cumulus Imports
 use cumulus_client_collator::service::CollatorService;
 use cumulus_client_consensus_common::ParachainBlockImport as TParachainBlockImport;
-use cumulus_client_consensus_proposer::Proposer;
+// Removed: cumulus_client_consensus_proposer::Proposer - ProposerFactory now directly implements ProposerInterface
```
la10736: Useless
```suggestion
```

relay-node/cli/Cargo.toml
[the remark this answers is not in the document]
la10736: It's not exposed anywhere, we can but the only ones that matters are runtime and node. We didn't care of this one and server before.

relay-node/service/Cargo.toml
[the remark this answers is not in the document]
la10736: Same as before. Anyway, I've not a strong opinion about that... we can use the same upstream version in  order to emphasize that are forks.

Any thought?

relay-node/service/upstream-differences.md
la10736: That's not the same like the before ones. That is exactly a report about the differences between the downstream version and the upstream one and should be maintained. The scope of this document is provide a documentation about the upstream differences and help future upstream alignment.

The `md` documents in `docs` folder was used in a claude prompt to do some tasks during the migrations: as I wrote before, I'll evolve them to generate useful generic skills.

So I prefer to not use the `_2512` suffix here because it's meant to reflect last upstream migration always in the future.

relay-node/test/runtime/Cargo.toml
```diff
@@ -173,6 +173,7 @@ runtime-benchmarks = [
 	"pallet-grandpa/runtime-benchmarks",
 	"pallet-indices/runtime-benchmarks",
 	"pallet-offences/runtime-benchmarks",
+ "pallet-session/runtime-benchmarks",
```
la10736: Fix this format

runtime/zkverify/src/genesis_config_presets.rs
```diff
@@ -50,6 +50,16 @@ fn session_keys(
...
+    let mut node_features = NodeFeatures::new();
+    node_features.resize(
+        node_features::FeatureIndex::FirstUnassigned as usize + 1,
+        false,
+    );
+    node_features.set(
+        node_features::FeatureIndex::CandidateReceiptV2 as u8 as usize,
+        true,
+    );
+
```
la10736, line 62: Double check WHY you don't need `ElasticScalingMVP` and `EnableAssignmentsV2` too.

For the first I guess is that the new node has only this behavior and so if you use this node you don't need to activate this node feature. For the second one I don't know.

verifiers/ultrahonk/src/lib.rs
```diff
@@ -18,7 +18,10 @@
 extern crate alloc;
 
 use alloc::{borrow::Cow, vec::Vec};
-use codec::{Decode, Encode, MaxEncodedLen};
+use codec::Decode;
+use codec::DecodeWithMemTracking;
+use codec::Encode;
+use codec::MaxEncodedLen;
```
la10736: nest them

runtime/migration_to_2512.md
```diff
@@ -0,0 +1,553 @@
...
+
+**Storage Version Change**: 15 → 16
+
+**Migration**: `pallet_staking::migrations::v16::MigrateV15ToV16<T>`
+
+**Purpose**: Migrates `DisabledValidators` from `Vec<u32>` to `Vec<(u32, OffenceSeverity)>` to track offense severity for re-enabling purposes.
+
+**Details**:
+- Uses max severity (`Perbill::from_percent(100)`) for all existing disabled validators
+- This means offenders before the migration will not be re-enabled this era unless there are other 100% offenders
```
la10736, line 347: Analyze the attack risk in case of "too long migration"

runtime/migration_to_2512.md
```diff
@@ -0,0 +1,553 @@
...
+);
+```
+
+### 2. pallet-session: v0 → v1
+
+**Storage Version Change**: 0 → 1
+
+**Migration**: `pallet_session::migrations::v1::MigrateV0ToV1<T, S>`
+
+**Purpose**: Migrates disabled validators storage to work with the new staking DisabledValidators format (with OffenceSeverity).
```
la10736, line 368: Analyze the attack risk in case of "too long migration": should be bounded by max validators, but check it

runtime/migration_to_2512.md
```diff
@@ -0,0 +1,553 @@
...
+
+**Details**:
+- The genesis config already enables the feature for new chains, but existing chains upgrading need this migration
+- Dispatches `configuration::Pallet::set_node_feature(Root, 3, true)` to schedule the change via the pallet's own `schedule_config_update` logic, avoiding direct manipulation of pallet-internal storage
+- The feature activates after a session delay (`cur+2`), which is the standard configuration change cadence
+- Uses `.expect()` on the dispatch result: `set_node_feature` with `RawOrigin::Root` and a valid feature index has no realistic failure path (the origin check always passes, and the consistency check won't reject a node feature toggle). If it somehow did fail, panicking is preferable to silently proceeding without the feature — that would break parachain block backing. Note: a panic in `on_runtime_upgrade` reverts the entire block, effectively preventing the new runtime from producing blocks; recovery would require a governance `system.setCode` with a fixed runtime
+
+**Location**: Defined in `runtime/zkverify/src/parachains.rs` (parachain migrations), composed into the runtime's `Migrations` tuple via `ParachainMigrations` in `lib.rs`
```
la10736: We removed this migration.

runtime/migration_to_2512.md
```diff
@@ -0,0 +1,553 @@
...
+The `slashes_go_to_treasury` test was simplified because the complex era transition simulation became unreliable with SDK 2512 changes. The new test directly verifies the `Slash` configuration works correctly without simulating the full staking flow.
+
+### 2. Balance Inspection
+
+Code that uses `Balances::balance()` expecting total balance (including staked amounts) must be updated to use `Balances::total_balance()` or `<Balances as Inspect<_>>::total_balance()`.
+
+### 3. Paratest Collator PeerId
+
+The paratest collator node (`paratest/node/src/service.rs`) was using `PeerId::random()` for the collator peer ID passed to the lookahead collator `AuraParams`. This prevented proper collator-validator communication because the advertised peer ID didn't match the actual network peer ID. Fixed by passing `network.local_peer_id()` through to `start_consensus()`.
+
```
la10736, line 459: Is this still true?

runtime/migration_to_2512.md
la10736, line 1: Please use claude (multi agent) to review this docs.

la10736 (commented):
Something should be addressed

runtime/zkverify/src/genesis_config_presets.rs
la10736: We double checked it. Only `ElasticScalingMVP` is interesting but both node and runtime from 2512 have just  this behavior,  that means 
1. We don't need to enable it in genesis at least we would generate a genesis and use it with the old node.... wired
2. This feature SHOULD be set on the old runtime BEFORE update the new node ... otherwise we'll brick  the chain.

relay-node/service/upstream-differences.md
drgora: Ok makes sense, we just need to remember to keep this up to date. I see we're going to have a `refine_patches` skill, so maybe we could add a reminder there?

docs/investigate_benchmark_issue.md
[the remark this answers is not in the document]
la10736: Moved in a skill and handle both debug and release

docs/investigate_benchmark_issue.md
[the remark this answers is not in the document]
la10736: Moved into a skill

relay-node/service/upstream-differences.md
la10736: Whenever  we will do the new upstream alignment we'll use this document to help this alignment ... I guess that will be natural upgrade it for the future migrations.

Anyway, where you propose to put this note?

drgora: I was suggesting to add a note to the `refine-patch` skill, which is most likely run whenever we update to a newer sdk.

.github/actions/cmd-in-docker/action.yml
```diff
@@ -25,6 +25,16 @@ inputs:
...
+  llvm:
+    description: 'Whether to install llvm apt package. If not provided llvm is not being installed: yes or no'
+    required: false
+    type: string
+    default: "no"
+  libclang-dev:
+    description: 'Whether to install libclang-dev apt package. If not provided libclang-dev is not being installed: yes or no'
+    required: false
+    type: string
+    default: "no"
```
ysibirski, line 37: I think we should better install this into base-images instead to save runtime time as an option like we did this for lld and cmake: HorizenLabs/ci-base-docker@4f6be15

ysibirski (changes requested):
See one comment

ysibirski (approved):
LGTM

PR #416, opened 2026-04-10
Small fixes on aggregate pallet

la10736:
- Reduce log level
- Un hold the domain owner funds when the manager remove a domain
- Return the hod funds to the submitter when the domain ownr cannot receive the funds.


95DDB (approved):

drgora (approved):
LGTM, thanks!

PR #420, opened 2026-04-30
Fix cargo audit: update rustls-webpki and triage new advisories

la10736:
## Summary
- Update `rustls-webpki` dependency in `Cargo.lock` to fix a known vulnerability
- Add new advisory triage entries to `.cargo/audit.toml` for advisories that cannot be resolved via dependency updates

## Test plan
- [ ] Verify `cargo audit` passes with the updated configuration
- [ ] Verify `cargo check` still compiles successfully


la10736:
Already bumped

PR #427, opened 2026-06-22
Implement Kimchi verifier

ozankaymak:


native/Cargo.toml
```diff
@@ -16,6 +16,10 @@ ark-bn254 = { version = "0.5.0", default-features = false }
 ark-bn254-ext = { git = "https://github.com/zkVerify/accelerated-bn-cryptography.git", default-features = false, tag = "v0.6.0" }
 ark-ec = { version = "0.5.0", default-features = false }
 ark-scale = { features = ["hazmat"], version = "0.0.13", default-features = false }
+bincode = { version = "1.3", optional = true }
```
95DDB: This dependency is declared and enabled under std, but nothing in native/src/ uses it. Dead dependency in the security-critical node crate — please remove both lines.

verifiers/kimchi/src/builtin_opening.rs
```diff
@@ -0,0 +1,232 @@
...
+    where
+        EFqSponge: FqSponge<Fq, Vesta, Fp, FULL_ROUNDS>,
+        RNG: RngCore + CryptoRng,
+    {
+        verify_with_native_srs_msm(srs, group_map, batch, rng)
+    }
+}
+
+#[allow(clippy::too_many_lines)]
+fn verify_with_native_srs_msm<EFqSponge, RNG>(
```
95DDB, line 113: This function seems to be a hand-copy of upstream `poly_commitment::ipa::OpeningProof::verify`. Please add a comment pinning the exact upstream tag/lines it mirrors and a "re-diff on every proof-systems bump" note — silent divergence here is a soundness bug.

95DDB: Actually, it would be nice, upstream, to factor the final MSM out of verify into an overridable backend, s.t. this copy can shrink/disappear and we reduce possible divergence surface, but definitely not a blocker

verifier-nostd-check/kimchi/src/lib.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2026, Horizen Labs, Inc.
```
95DDB: Can you remove all the changes here please and maybe keep them locally ? This crate was inteded to be used locally and never pushed into

verifiers/kimchi/src/verifier_should.rs
```diff
@@ -0,0 +1,1088 @@
...
+    #[test]
+    fn foreign_field_mul_fixture_is_accepted() {
+        fixture_is_accepted(
+            include_bytes!("resources/foreign_field_mul/proof.bin"),
+            include_bytes!("resources/foreign_field_mul/verifier_index.bin"),
+            Vec::new(),
+        );
+    }
+
+    fn accelerated_opening_matches_upstream(raw_proof: Vec<u8>, raw_vk: Vec<u8>, raw_pubs: Pubs) {
```
95DDB, line 372: Nice differential test against upstream verify ! It only covers the accept direction on valid fixtures, though — the reject cases run the accelerated path only. To catch drift that makes the copy accept what upstream rejects, run the mutated/reject fixtures through upstream verify too and assert both reject, please

native/src/accelerated_pasta/vesta.rs
```diff
@@ -0,0 +1,920 @@
...
+use alloc::vec::Vec;
+use mina_curves::pasta::{Fp, ProjectiveVesta, Vesta};
+use sp_runtime_interface::pass_by::{AllocateAndReturnByCodec, PassFatPointerAndRead};
+use sp_runtime_interface::runtime_interface;
+
+use crate::accelerated_bn::utils;
+#[cfg(feature = "std")]
+use crate::VerifyError;
+
+const VESTA16_SRS_SIZE: usize = 1 << 16;
```
95DDB: Just a note here: because the SRS size (2^16), supported domain range, and parameter digests are baked into native, future Kimchi evolution (bigger SRS, new profile/curve) needs a coordinated node upgrade + host-function versioning, not just a runtime upgrade. This is always a heavy lift on our side, and not actionable in a relatively short time-frame. So if one of those are ever likely to change, I would suggest to re-discuss and re-design what has been baked into the native module here

95DDB (changes requested):
Very very excellent work, thank you very much !
I've left some comments, but I'll let @drgora do a more in-depth review.
Failing CI is probably on our side as we're in the process of upgrading the network, but will let you know.
A big IF that is still open revolves around benchmarks: we'd had to execute them on our reference machine and make sure the verification time doesn't get significantly worse compared to your results (executed on a Macbook it seems)

Cargo.toml
```diff
@@ -372,6 +374,9 @@ cumulus-relay-chain-rpc-interface = { version = "0.28.0" }
 substrate-rpc-client = { version = "0.54.0" }
 
 [patch.crates-io]
+# Kimchi pulls `bcs` through `o1-utils`; the crates.io release enables
+# `serde/std`, which breaks Substrate runtime WASM with duplicate `panic_impl`.
+bcs = { git = "https://github.com/o1-labs/bcs", rev = "d9c996b0" }
```
drgora, line 393: I see from the upstream activity that you guys are already close to having this merged, and the bcs developers seem quite responsive. Perhaps you could give another try, and hopefully remove this patch?

verifier-nostd-check/maybe-fail/Cargo.toml
```diff
@@ -11,4 +11,4 @@ bit-vec = { version = "0.8.0", default-features = false, features = ["std"] }
 
 [features]
 fail = []
-default = []
\ No newline at end of file
+default = []
```
drgora: Can you revert this and the other modifications in `verifier-nostd-check`? I believe it helps removing some noise from this huge PR

verifiers/kimchi/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+ark-ff = { version = "0.5", default-features = false }
+ark-poly = { version = "0.5", default-features = false }
+ark-serialize = { version = "0.5", default-features = false }
+bincode = { version = "2", default-features = false, features = ["alloc", "serde"] }
+codec = { workspace = true }
+frame-benchmarking = { workspace = true, optional = true }
+frame-support = { workspace = true }
+frame-system = { workspace = true }
+hex-literal = { workspace = true, optional = true }
+kimchi = { git = "https://github.com/o1-labs/proof-systems", tag = "0.7.0", default-features = false }
```
drgora: Can you align with the other verifiers and move this and similar imports to the root `Cargo.toml`, and import with `workspace = true` here?

native/src/accelerated_pasta/vesta.rs
drgora: I couldn't agree more on this, and would like to at least have an explicit ack before merging this PR

relay-node/service/src/lib.rs
```diff
@@ -348,11 +351,29 @@ pub fn open_database(db_source: &DatabaseSource) -> Result<Arc<dyn Database>, Er
     Ok(parachains_db)
 }
 
+/// Prepare all consensus-supported native verifier parameters before block execution.
+pub fn prewarm_native_verifier_parameters() -> Result<(), Error> {
+    const KIMCHI_DOMAIN_LOG2_SIZES: &[u8] =
```
drgora: This *must* cover all the possible domain sizes allowed by the verifier. We cannot afford a lazy init on block validation, right.
If so, could you either add some checks to make sure that we cover the entire domain (without holes) during init, or maybe change this function to implicitly do that?

native/Cargo.toml
drgora: Probably a good time to bump this to version `0.4.0`

native/src/accelerated_bn/mod.rs
```diff
@@ -17,7 +17,7 @@
 
 pub mod bn254;
 
-mod utils;
+pub(crate) mod utils;
```
drgora: Now that we have another module using it, it might be cleaner to raise the utils module one level (and perhaps rename it to something like `arkworks_utils`, so that we do not need to change this visibility, and let `accelerated_pasta` not depend on `accelerated_bn`

utils/kimchi-fixture-generator/src/main.rs
```diff
@@ -0,0 +1,332 @@
+// Copyright 2026, Horizen Labs, Inc.
```
drgora, line 1: I believe the `utils` top level directory might be a better fit for this generator? In the scripts dir we do not have any Rust code. And, also to make it consistent with the rest, add it to the `exclude` field of the top level `Cargo.toml`.

Also, could you add some documentation on what this tool is for and how to use it?

verifiers/kimchi/src/profile.rs
```diff
@@ -0,0 +1,483 @@
...
+///
+/// Generic Kimchi custom gates and lookup arguments are supported. Recursive
+/// accumulators remain a Pickles concern. The chunk and public-input caps are
+/// the benchmarked block-work envelope.
+pub(crate) struct Vesta16;
+
+impl KimchiProfile for Vesta16 {
+    const MAX_DECODED_PROOF_BYTES: usize = 262_144;
+    const MAX_DECODED_VK_BYTES: usize = 65_536;
+    const MAX_CHUNKS: usize = 4;
```
drgora, line 69: It seems that this (and maybe few others?) parameter must be kept aligned with the constraints in the native crate. Could you add some assertions to make sure that the configuration is consistent?
Similarly in the `runtime/zkverify/src/lib.rs`, can we make sure that the parameters set are compatible with the contraints? You can find similar `const_assert!` for example

relay-node/service/src/builder/mod.rs
```diff
@@ -197,6 +197,7 @@ where
         mut config: Configuration,
         params: NewFullParams<OverseerGenerator>,
     ) -> Result<ServiceBuilder<OverseerGenerator, Network>, Error> {
+        crate::prewarm_native_verifier_parameters()?;
```
drgora: I measured this overhead on my 16 cores cpu, and it takes ~240s. I believe that is too much in general, and even more in the development context.

I would like to explore other solutions that do not incur this overhead on every init: for example, how about caching the precomputed parameters to the local storage? You could trigger the computation only if the precomputed data is missing or fails verification.
To me this is a simple yet effective solution that avoid other complexities or packaging burdens, but I'm open to discussing this point.

drgora (changes requested):
Thank you guys for this PR, this is an impressive job!
I believe this needs a bit of refinement for being ready for merge, but the direction is very promising.
Also, I fully agree with @95DDB on the benchmarks: we should recompute them on the reference machine, but I'm okay with doing this in a follow up PR.

relay-node/service/src/builder/mod.rs
ozankaymak: Will do that, thank you

native/src/accelerated_pasta/vesta.rs
ozankaymak: Explicitly acknowledged. O1 Labs confirmed that the SRS size will not change in the current hard fork and is not expected to change in the next. Will also document in detail.

PR #429, opened 2026-07-06
ci: rustdoc redirect + prod-only docs, and coverage disk fix

ysibirski:
## Problems

**1. Docs landing page 404s.** The rustdoc job writes a meta-refresh `index.html` that redirects to `zkv_node/index.html`, but there is no `zkv_node` crate — the main binary is `zkv-relay`. So `https://zkverify.github.io/zkVerify/` loads, then bounces to a 404. Confirmed live:

| URL | Status |
|-----|--------|
| `https://zkverify.github.io/zkVerify/` | 200 (redirects…) |
| `.../zkv_node/index.html` | **404** (redirect target) |
| `.../zkv_relay/index.html` | 200 ✅ |

**2. Rustdoc deploy fails on non-prod tags.** `CI-tag-orchestrator.yml` ran the Rustdoc job with `deploy: 'yes'` for every tag it triggers on (prod `-[0-9]+`, `-rc[0-9]+`, and test `-[a-zA-Z]*`). But the `github-pages` environment only allows prod-format tags (`v[0-9]*.[0-9]*.[0-9]*-[0-9]*`), so RC/test tags fail with:

```
Tag "v2.0.0-rc1" is not allowed to deploy to github-pages due to environment protection rules.
```

**3. Coverage job fails with `No space left on device` (os error 28).** `cargo llvm-cov --all-features --workspace` instruments the entire Substrate workspace in the `dev` profile, which defaults to `debug = 2` (full DWARF) — the largest component of the target tree. The WarpBuild runner has a fixed 150 GB disk (every Linux tier; [not resizable](https://www.warpbuild.com/docs/ci/cloud-runners)), so the fix must reduce what the build writes.

## Changes

- **`CI-rustdoc.yml`** — redirect target `zkv_node/index.html` → `zkv_relay/index.html`.
- **`CI-release.yml`** — expose the existing `check-requirements` job output `is-a-prod-release` as a `workflow_call` output, so callers can reuse the canonical PROD determination from `check_release_requirements.sh` (no duplicated tag-format logic).
- **`CI-tag-orchestrator.yml`** — `rustdoc-job` now `needs: [release]` and runs only when `is-a-prod-release == 'true'`, so docs deploy for production tags only.
- **`CI-coverage.yml`** — build coverage with `CARGO_PROFILE_DEV_DEBUG=0`. LLVM source-based coverage embeds its line/region mapping in `__llvm_covmap` and does not need DWARF, so dropping debug info reclaims the disk with no loss of coverage. Scoped via the profile env var, not `RUSTFLAGS` (which `cargo-llvm-cov` manages for `-C instrument-coverage`).

## Notes

- Gating docs to prod tags matches the `github-pages` environment guardrail already in place, and avoids overwriting released docs with pre-release (RC) content.
- `rustdoc-job` depends on `release`, so docs deploy after a successful release rather than in parallel — intentional.
- The coverage fix takes full effect once merged to the default branch (the restored cache shrinks after the first default-branch run). If a run still approaches 150 GB, the follow-up is to shard `--workspace` coverage across parallel jobs and merge the `lcov.info` files.

### DIFF
--- a/.github/workflows/CI-coverage.yml
+++ b/.github/workflows/CI-coverage.yml
@@ -18,7 +18,7 @@ jobs:
         uses: ./.github/actions/cmd-in-docker
         with:
           command: >-
-            /bin/bash -c 'set -o pipefail && cargo llvm-cov clean --workspace && find . -name \"*.profraw\" -delete
+            /bin/bash -c 'set -o pipefail && export CARGO_PROFILE_DEV_DEBUG=0 && cargo llvm-cov clean --workspace && find . -name \"*.profraw\" -delete
             && cargo llvm-cov --no-rustc-wrapper --all-features --workspace --exclude zkv-relay --exclude paratest-node --exclude test-service --lcov --output-path lcov.info
             && cargo llvm-cov report --json --output-path coverage_report.json --summary-only && cargo llvm-cov report | tee coverage_summary.txt'
           use_cache: 'yes'
--- a/.github/workflows/CI-release.yml
+++ b/.github/workflows/CI-release.yml
@@ -18,6 +18,10 @@ on:
         required: true
         type: boolean
         default: false
+    outputs:
+      is-a-prod-release:
+        description: "Whether the tag is a production release"
+        value: ${{ jobs.check-requirements.outputs.is-a-prod-release }}
     secrets:
       DOCKER_HUB_USERNAME:
         required: true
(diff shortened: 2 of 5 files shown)


drgora (commented):
LGTM, thank!

drgora (approved):

davehorizen (approved):

drgora (approved):

